PR #1, opened 2017-06-14
initial commit.  Created a basic webpack.config.js file, some basic .…

rarebreed:
…babelrc settings, and some notes on the build

Here's a basic webpack 2 build setup.  It has some babel settings already and the package.json has a few things installed as dev dependencies, mainly webpack itself and some babel node modules.

I left the directory structure mostly free.  I just assumed a src/components folder for now (assuming this is where react components would live).  The entry point for webpack to build the dependency graph is src/index.js.

PR #2, opened 2017-06-19
Initial sample page

dperpeet:
follow-up to #1


webpack.config.js
```diff
@@ -1,18 +1,113 @@
...
+/*
+ * Note that we're avoiding the use of path.join as webpack and nodejs
+ * want relative paths that start with ./ explicitly.
+ *
+ * In addition we mimic the VPATH style functionality of GNU Makefile
+ * where we first check builddir, and then srcdir.
+ */
+
+function vpath(/* ... */) {
+    var filename = Array.prototype.join.call(arguments, path.sep);
```
stefwalter: I have not seen ES6 in use in webpack.config.js ... which is a Turing complete configuration file :S  My opinion: Tread cautiously with ES6 usage there.

src/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "version": "0.1",
```
rarebreed: Do we want to use just a version with two elements or a semantic versioning with major, minor, patch?

Makefile
```diff
@@ -0,0 +1,21 @@
+all:
+	npm run build
```
rarebreed: Do you want to just use npm?  I'm ok with that since it turns out yarn, while a drop in replacement for npm with respect to module management and package.json compliancy, does have a slightly different CLI usage (ie yarn install isn't the same as npm install...that's yarn add).

webpack.config.js
dperpeet: > I have not seen ES6 in use in webpack.config.js ... which is a Turing complete configuration file :S My opinion: Tread cautiously with ES6 usage there.

Agreed. Let's be careful.

Makefile
dperpeet: I think for make that would be good, since yarn isn't available on Fedora.

PR #3, opened 2017-07-07
Add Vagrantfile

larskarlitski:
A simple Vagrant VM based on fedora 25 cloud. It syncs `dist` to
`/usr/local/share/cockpit/<dirname>` on the VM.

Also decrease the minimum required version in `manifest.json` to match
the one that's currently in fedora 25.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,8 +2,12 @@ Vagrant.configure(2) do |config|
     config.vm.box = "fedora/25-cloud-base"
     config.vm.network "forwarded_port", guest: 9090, host: 9090
 
+    if Dir.glob("dist/*").length == 0
+      config.vm.post_up_message = "NOTE: Distribution directory is empty. Run `make` to see your module show up in cockpit"
+    end
+
     config.vm.synced_folder ".", "/vagrant", disabled: true
-    config.vm.synced_folder "dist/", "/usr/local/share/cockpit/" + File.basename(Dir.pwd)
+    config.vm.synced_folder "dist/", "/usr/local/share/cockpit/" + File.basename(Dir.pwd), create: true
 
     config.vm.provider "libvirt" do |libvirt|
         libvirt.memory = 1024
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,26 @@
+Vagrant.configure(2) do |config|
+    config.vm.box = "fedora/25-cloud-base"
+    config.vm.network "forwarded_port", guest: 9090, host: 9090
+
+    config.vm.synced_folder ".", "/vagrant", disabled: true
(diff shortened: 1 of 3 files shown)


src/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
     "version": "0.1",
     "requires": {
-        "cockpit": "138"
+        "cockpit": "137"
```
dperpeet: Not quite sure about this, since we changed quite a bit around 138

But it should be ok for the little while until fedora 26 is available as a good base.

Vagrantfile
```diff
@@ -0,0 +1,26 @@
+Vagrant.configure(2) do |config|
+    config.vm.box = "fedora/25-cloud-base"
+    config.vm.network "forwarded_port", guest: 9090, host: 9090
```
dperpeet: should we pick a different port on the host, in case it also has cockpit running?

src/manifest.json
larskarlitski: Yeah I'm not to sure about that either. The alternative would be to use cockpit from the copr for this release.

Vagrantfile
larskarlitski: Hm, good question. I picked this one because that's where cockpit usually listens on. I'm afraid another port could confuse people, or do we have another "development port" convention?

src/manifest.json
dperpeet: well, we can always change this later, let's keep it simple for now

Vagrantfile
dperpeet: Good point. Let's stick with the default, it's not that difficult to change the Vagrantfile if necessary.

Vagrantfile
```diff
@@ -0,0 +1,26 @@
...
+
+    config.vm.synced_folder ".", "/vagrant", disabled: true
+    config.vm.synced_folder "dist/", "/usr/local/share/cockpit/" + File.basename(Dir.pwd)
```
dperpeet: Can we auto-create these directories if necessary? On a clean repo I get:

```
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The host path of the shared folder is missing: dist/
```

dperpeet: As discussed on irc, I propose just creating this directory and printing a brief warning along the lines of `Distribution directory is currently empty, please call make`

larskarlitski: Indeed, that makes it easier to figure out when the module doesn't show up in the VM. Thanks for the suggestion! Pushed another commit that adds the message and creates the directory when it doesn't exist.

dperpeet:
Works:
```
==> default: Machine 'default' has a post `vagrant up` message. This is a message
==> default: from the creator of the Vagrantfile, and not from Vagrant itself:
==> default: 
==> default: NOTE: Distribution directory is empty. Run `make` to see your module show up in cockpit
```

PR #4, opened 2017-07-11
[WIP] Proof of concept for a simple unit test

rarebreed:
Here's an idea to be able to do React component unit testing.  It uses the Tape testing framework which is very simple and doesn't do anything with the global environment (unlike mocha or jasmine).  It also can run tests inside the browser itself, which would allow for importing of real react components into the DOM and doing tests that way (as opposed to just selenium or phantom).  I also included dependencies for the flow language, including what's needed by babel.

I don't know if there is a better way to do this, but you have to do three things for the test:
- yarn run flow-src      # to get babel to transpile the flow and jsx code to /lib
- yarn run flow-tests   # to get the tests transpiled to /testlib
- node testlib/concept.test.js   # to actually run the test

I'm welcome to a better way to do this.  It took me a lot of hunting to figure out how to get the tests transpiled.  I know webpack will only use babel for anything touched by the entry point (not to mention it will bundle everything inside index.js)


src/components/test-component.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react';
+
+function TestComponent(props) {
```
rarebreed: Hmmm, weird, I forgot to put "export" here.  And yet, in the test code, I was able to import it.  I didn't think es6 modules would allow that.

rarebreed:
Closing this out, as new work will be continued from a new repo:
https://github.com/candlepin/subscription-manager

Work will be on the feature/cockpit branch

PR #5, opened 2017-07-25
Updated example and more build options

dperpeet:


package.json
```diff
@@ -23,7 +24,6 @@
     "webpack": "^2.6.1"
   },
   "dependencies": {
-    "react": "^15.6.0",
-    "react-dom": "^15.6.0"
+    "react-lite": "0.15.30"
```
dperpeet: I chose this version to match what the cockpit main repo [is using](https://github.com/cockpit-project/cockpit/blob/ea5507dfd4cd167ea9a38c6ec31bc08a26387797/package.json#L26). I don't think the exact version matters so much, since we only use basic stuff.

src/index.es6
```diff
@@ -0,0 +1,39 @@
+/*jshint esversion: 6 */
+/*
+ * This file is part of Cockpit.
```
dperpeet: I think this is more of a sample project / starter kit. Should we get rid of this line altogether?

Makefile
```diff
@@ -1,21 +1,40 @@
...
 
 install: all
-	mkdir -p /usr/share/cockpit
-	cp -r dist/ /usr/share/cockpit/subscription-manager
+	mkdir -p /usr/share/cockpit/subscription-manager
+	cp -r dist/* /usr/share/cockpit/subscription-manager
 
-srpm: clean all
+dist-gzip: NODE_ENV=production
+dist-gzip: clean all
```
dperpeet: No, this calls `clean` and `all` with the environment variable set to production. See [target-specific variable values](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html)

package.json
dperpeet: After some discussion on irc we came to the conclusion that this is "locked down" already

src/index.es6
larskarlitski: Yeah, I agree.

Makefile
larskarlitski: Indeed, sorry.

dperpeet:
Thanks for reviewing, I addressed the issues and pushed new commits.

> Vagrant works like before, but it's a bit weird to have two things that watch directories: webpack and vagrant. (Obviously that's unrelated to this work.)

It can seem a bit strange, but I think this caters to completely different workflows. I for one don't use vagrant at all, so the webpack workflow is very convenient to me.

larskarlitski:
> Thanks for reviewing, I addressed the issues and pushed new commits.

Thanks!

PR #8, opened 2017-08-22
Improve source package slightly

mvollmer:
This moves us one step closer to a normal source package.  The sources
are now unpacked and we run "make install".  We still don't build
since we don't have all the necessary tools.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 node_modules/
 dist/
 /.vagrant
+/webpack.inputs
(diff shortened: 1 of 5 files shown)


larskarlitski:
Cool, thanks!

Before I merge, what does `webpack-with-stats` do?

mvollmer:
> Before I merge, what does webpack-with-stats do?

It writes the `webpack.inputs` file that is then used to make the dist tarball.

PR #11, opened 2017-10-18
Dist using `git ls-files`

larskarlitski:
This makes sure we always include all files in the repository when
disting.

Obsoletes webpack-with-stats, which had a bug anyway: it wasn't
reporting jslint errors.


mvollmer:
> Obsoletes webpack-with-stats

I am not that this is true (but I might be confused): We want to put files into the source rpm that are not in git but have been downloaded by `npm install`.  However, we don't want to include all of `node_modules`, but only the files that are included in `dist/` by webpack.

martinpitt:
My gut feeling is that we shouldn't make the starter kit more complicated than necessary, and use the JS tools as much as possible. If we distribute the webpack-generated `dist/` and `package-lock.json` then it's easy enough to build without patches (as you don't need to run webpack or anything), and possible to reproduce exactly if you do need patches (through the pinpointed versions in package-lock). Or the user of the starter kit can decide to dist the entire `node_modules/` for their project.

Note that the webpack-with-stats solution also isn't ideal - for example, it skips licenses, package.json etc.

mvollmer:
> If we distribute the webpack-generated dist/ and package-lock.json

That's an excellent point.  Should we do the same for Cockpit itself?

martinpitt:
> Should we do the same for Cockpit itself?

I wouldn't mind doing that, especially as we already build and release the -cache tarball.

PR #16, opened 2018-03-08
Add integration test using Cockpit test API

martinpitt:
And some build system fixes.

Tests here won't run until starter-kit gets enabled in Cockpit CI. I'll do that next

 - [x] enable starter-kit tests in Cockpit CI


martinpitt:
I triggered the two requests manually, but they won't be served yet.

PR #26, opened 2018-06-21
Add scss example

larskarlitski:


src/app.jsx
```diff
@@ -36,9 +37,9 @@ export class Application extends React.Component {
         return (
             <div className="container-fluid">
                 <h2>Starter Kit</h2>
-                <div>
-                    <span>{ cockpit.format(_("Running on $0"), this.state.hostname) }</span>
```
martinpitt: The tests check (and now fail on) this span, so you need to adjust the tests accordingly.

PR #36, opened 2018-07-17
Various build system and srpm fixes

martinpitt:
With these plus a [cockpituous fix](https://github.com/cockpit-project/cockpituous/pull/194) one now can generate working upstream release tarballs, srpm, and build the srpm in mock (thus no network access or missing dependencies) to generate a working rpm. Tested like this:
```
git clean -ffdx
RELEASE_SOURCE=$PWD/_release/source  ~/upstream/cockpituous/release/release-source 
RELEASE_SOURCE=$PWD/_release/source  RELEASE_SPEC=cockpit-starter-kit.spec RELEASE_SRPM=_release/srpm ~/upstream/cockpituous/release/release-srpm -V
mock _release/cockpit-starter-kit-999-1.fc28.src.rpm
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,10 @@ export TEST_OS
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 # one example directory from `npm install` to check if that already ran
 NODE_MODULES_TEST=node_modules/po2json
+# one example file in dist/ from webpack to check if that already ran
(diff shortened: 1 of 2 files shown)


martinpitt:
I pushed another fix that caused an RPM build failure for @sabbaka in the session-recording project.

martinpitt:
I debugged another corner case with @sabbaka, testing a fix.

PR #40, opened 2018-08-28
Various build fixes and improvements

martinpitt:


martinpitt:
@larskarlitski : Good point, thanks. Changed to https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Good_Licenses , which is the direct link to the table.

PR #46, opened 2018-09-06
test: Really clean up test VM on image preparation

martinpitt:
The previous fix (commit 6e05f5b4837) only cleaned
test/images/$(TEST_OS), which is just a symlink to $(TEST_OS).qcow.
Clean the actual image as well.


martinpitt:
What a blunder, sorry about that!

PR #50, opened 2018-10-28
WIP: Move test to cypress.io

martinpitt:
This is mostly just a PoC to see if our CI infrastructure is able to run
cypress.io tests.
    
Does not yet cover changing language, this is a bit tricky with cypress.

 - [x] Add Xvfb to cockpit/tests container: https://github.com/cockpit-project/cockpituous/pull/221
 - [x] Fix "500 Parse Error" in Cockpit when loading manifest.{js,json}: https://github.com/cockpit-project/cockpit/pull/10499


martinpitt:
The [previous test run](https://fedorapeople.org/groups/cockpit/logs/pull-50-20181029-081717-85e37345-cockpit-fedora-28/log) worked, but does not yet include the language testing part. Today I (mostly) debugged why this fails and causes a "500 Parse Error" in cypress. It seems this does not get along with sending the manifest JS or JSON with chunked encoding. We also don't really need that, as we already know the content length. I'll send a Cockpit PR.

martinpitt:
I haven't had time for this for half a year now, and in the meantime we improved our Python test API. I'm closing this for now, I can dig this out again when needed.

PR #94, opened 2019-01-27
Watch by default the hostname.

bmustiata:
Watch the hostname by default.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -28,6 +28,11 @@ class TestApplication(testlib.MachineCase):
         b.wait_present(".container-fluid p")
         b.wait_text(".container-fluid p", "Running on " + hostname)
 
+        # change current hostname
+        m.execute("echo new-%s > /etc/hostname" % hostname)
+        # verify new hostname name
+        b.wait_text(".container-fluid p", "Running on new-" + hostname)
+
         # change language to German
         b.switch_to_top()
         b.click("#content-user-name")
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -28,10 +28,9 @@ export class Application extends React.Component {
         super();
         this.state = { 'hostname': _("Unknown") };
 
-        cockpit.file('/etc/hostname').read()
-                .done((content) => {
-                    this.setState({ 'hostname': content.trim() });
-                });
+        cockpit.file('/etc/hostname').watch(content => {
+            this.setState({ 'hostname': content.trim() });
+        });
     }
 
     render() {


bmustiata:
For this one, you can thank your colleagues at devconf Brno, for spotting
it :)

I will try to create the integration test, since it seems quite
straightforward.

test/check-application
```diff
@@ -15,37 +15,64 @@ import testlib
...
+        # And the page label (from js) should be translated
+        browser.wait_in_text(".container-fluid p", "Läuft auf")
+
+    def testChangingHostnameFileWatching(self):
+        # Given on a cockpit instance
+        browser = self.browser
+        machine = self.machine
+        initial_hostname = machine.execute("cat /etc/hostname").strip()
+
+        self.login_and_go("/starter-kit")
```
martinpitt: This would make sense as part of `goToStarterKit()`, as it's the *only* piece of code that actually "goes" somewhere.

test/check-application
```diff
@@ -15,37 +15,64 @@ import testlib
...
+        # Then the hostname is also updated in the starter page
+        browser.wait_text(".container-fluid p", "Running on " + new_hostname)
+        assert new_hostname == "new_" + initial_hostname
+
+    def goToStarterKit(self, browser, machine):
         # verify expected heading
-        b.wait_present(".container-fluid h2")
-        b.wait_text(".container-fluid h2", "Starter Kit")
+        browser.wait_present(".container-fluid h2")
+        browser.wait_text(".container-fluid h2", "Starter Kit")
```
martinpitt: Also, these two lines really need to stay together with the `login_and_go`.

test/check-application
```diff
@@ -15,37 +15,64 @@ import testlib
...
-        # menu label (from manifest) should be translated
-        b.wait_text("#host-apps a[href='/starter-kit']", "Bausatz")
-
-        b.go("/starter-kit")
-        b.enter_page("/starter-kit")
-        # page label (from js) should be translated
-        b.wait_in_text(".container-fluid p", "Läuft auf")
+        hostname = machine.execute("hostname").strip()
+        browser.wait_present(".container-fluid p")
+        browser.wait_text(".container-fluid p", "Running on " + hostname)
```
martinpitt: This is not conceptually a part of "go to the starter kit page", it's an independent check.

test/check-application
```diff
@@ -15,37 +15,64 @@ import testlib
...
+
+        # When I change the hostname file on the underlying machine
+        machine.execute("echo new_$(hostname) > /etc/hostname").strip()
+        new_hostname = machine.execute("cat /etc/hostname").strip()
+
+        # Then the hostname is also updated in the starter page
+        browser.wait_text(".container-fluid p", "Running on " + new_hostname)
+        assert new_hostname == "new_" + initial_hostname
+
+    def goToStarterKit(self, browser, machine):
```
martinpitt: There's no need to pass browser and machine, `self` has these.

test/check-application
```diff
@@ -15,37 +15,64 @@ import testlib
 
 class TestApplication(testlib.MachineCase):
     def testBasic(self):
-        b = self.browser
-        m = self.machine
+        """
+        Tests if changing the language works as expected.
+        """
```
martinpitt: I deliberately did not split the test, as each new test has a considerable setup cost. But I'm okay with doing it as a way of demonstrating how to organize larger test suites. But then let's also rename this to `testTranslation`.

bmustiata:
For some reason now, on my machine, starting today, I kept getting in the test a 401 error (`TestApplication-testBasic-centos-7-127.0.0.2-2201-FAIL.js.log`):

```
cdp: {'source': 'network', 'level': 'error', 'text': 'Failed to load resource: the server responded with a status of 401 (Authentication failed)', 'timestamp': 1548705758942.69, 'url': 'http://127.0.0.2:9091/cockpit/login', 'networkRequestId': '1000006248.8'}
```

I see that travis passes, which makes it even weirder.

I tried deleting the `tmp` folder, that apparently has the image, but it still doesn't work. Every test run takes over 2 minutes on my machine, so I'm not going to try anymore to figure out what's going on. I lost ~1 hour this evening just to implement these PR changes.

martinpitt:
BTW, to quickly re-run the test if you *only* changed the test, but not the code, you can just call `test/check-application` directly. Possibly with `-t` and `-v` output for more verbosity what it's doing.

I force-pushed a little cleanup here to get this out of the way, this looks good and works now. Thanks!

bmustiata:
I will do another PR for the test changes.

Personally I'm a big fan of BDD, and I think the `Given`/`When`/`Then` structure is what makes most sense. Now a When or Given can still actually check if they executed successfully themselves, especially if they're composed from multiple smaller actions. So I'd like to keep the post conditions in the actions - I've noticed you prefer them as separate steps altogether.

https://github.com/cockpit-project/starter-kit/pull/94#discussion_r251398410

Would that be ok?

martinpitt:
@bmustiata : It feels odd and unrelated to have a test setup function (go to the right page) also have check *some* functionality of that page. But we can discuss on the new PR. Thanks!

PR #147, opened 2019-03-04
tasks: Minimize races filing issues for bots

martinpitt:
We don't need to file or check for these issues on every iteration
through the .tasks script. These will count as "work done" and thus skip
the sleep (even though the "work" lasts only a split-second), and are
prone to creating duplicates.

Similar to https://github.com/cockpit-project/cockpit/commit/af2e934a

### DIFF
--- a/.tasks
+++ b/.tasks
@@ -1,7 +1,16 @@
 #!/bin/sh
 
-# Open issues for things that need doing on a regular basis
-bots/npm-trigger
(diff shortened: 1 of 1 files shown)


martinpitt:
Tested with `. .tasks </dev/null` and it doesn't do anything in about 9 out of 10 cases, and in the tenth case it outputs the expected issues.

PR #232, opened 2019-10-01
Makefile: Don't clobber an existing bots checkout

martinpitt:
Commit 95b2aff0 was a thinko -- for our CI we *don't* want our test to
clobber a pre-existing bots/ checkout, as we often use this to run tests
against an updated image or to validate a changes to the bots project.

On developer machines, bots may also be a symlink to an actual bots
directory in development, so don't clobber that.

 - Same fix in c-composer: https://github.com/weldr/cockpit-composer/pull/794
 - Same fix in cockpit-podman: https://github.com/cockpit-project/cockpit-podman/pull/213
 - Same fix in cockpit-ostree: https://github.com/cockpit-project/cockpit-ostree/pull/29

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,11 +136,7 @@ check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
 # checkout Cockpit's bots for standard test VM images and API to launch them
 # must be from master, as only that has current and existing images; but testvm.py API is stable
 bots:
-	if [ ! -d bots ]; then \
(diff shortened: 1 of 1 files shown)


martinpitt:
The [test log](https://209.132.184.41:8493/logs/pull-232-20191001-204545-d46ebbb7-cockpit-project-starter-kit--centos-7/log) proves that it does not check out bots any more (make-checkout already did that).

PR #233, opened 2019-10-02
Support CI testing against a bots project PR

martinpitt:
* If `$COCKPIT_BOTS_REF` is set, check out that bots version instead of
   master.

 * Use git cache in $XDG_CACHE_HOME if available. Our CI uses that to
   save downloads, and it does not get in the way for local developers.

 * Stop making "bots" a phony target, and drop the now unnecessary
   `[ -d bots ]` check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,8 +135,11 @@ check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
 
 # checkout Cockpit's bots for standard test VM images and API to launch them
 # must be from master, as only that has current and existing images; but testvm.py API is stable
+# support CI testing against a bots change
 bots:
-	[ -d bots ] || git clone --depth=1 https://github.com/cockpit-project/bots.git
+	git clone --quiet --reference-if-able $${XDG_CACHE_HOME:-$$HOME/.cache}/cockpit-project/bots https://github.com/cockpit-project/bots.git
+	if [ -n "$$COCKPIT_BOTS_REF" ]; then git -C bots fetch --quiet --depth=1 origin "$$COCKPIT_BOTS_REF"; git -C bots checkout --quiet FETCH_HEAD; fi
+	@echo "checked out bots/ ref $$(git -C bots rev-parse HEAD)"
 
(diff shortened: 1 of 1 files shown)


martinpitt:
This is currently done by `make-checkout --bots-ref`, but we want to get rid of that. It's too magic, and we already ran into troubles with that yesterday (see PR #232), and we want to simplify what the bots do as much as possible.

So the plan is:
 1. Land this change in all projects that use our CI
 2. Drop the --bots-ref stuff from make-checkout and change tests-scan to set `COCKPIT_BOTS_REF` instead of `--bots-ref`
 3. Clean up the rev vs. revision confusion in tests-scan vs. make-checkout

martinpitt:
Tested like this:

 * `git clean -ffdx; make bots` → bots is at master
 * `git clean -ffdx; COCKPIT_BOTS_REF=pull/32/head  make bots` → `git -C bots shortlog` confirms that the head is "images: Shrink windows-10 image and install QRes"
 * A subsequent `make bots` says "make: 'bots' is up to date." and does not clobber the existing checkout.

Makefile
```diff
@@ -135,8 +135,11 @@ check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
 
 # checkout Cockpit's bots for standard test VM images and API to launch them
 # must be from master, as only that has current and existing images; but testvm.py API is stable
+# support CI testing against a bots change
 bots:
-	[ -d bots ] || git clone --depth=1 https://github.com/cockpit-project/bots.git
+	git clone --quiet --reference-if-able $${XDG_CACHE_HOME:-/nonexisting}/cockpit-project/bots https://github.com/cockpit-project/bots.git
```
Gundersanne: I'm fine with this as-is, but according to https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html the default value for XDG_CACHE_HOME should be ~/.cache. 

It's not that important and I know I didn't do this in other places in cockpit myself, but I want to make a PR to bots to change this.

martinpitt: Good point, thanks! Updated. Let's do this correctly.

PR #239, opened 2019-10-16
Add eslint rules for React hooks

martinpitt:
These make components with simple state (only a few variables) easier to
read and maintain. See https://reactjs.org/docs/hooks-intro.html for
details. Hook are available since React 16.8, and we already depend on 16.10.

We can't use hooks in our actual code, as our only `Application`
component needs a constructor. But this enables the ESLint plugin to
guide developers to the right path if they use hooks.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,7 +13,7 @@
         },
         "sourceType": "module"
     },
-    "plugins": ["flowtype", "react"],
+    "plugins": ["flowtype", "react", "react-hooks"],
     "rules": {
         "indent": ["error", 4,
             {
@@ -27,6 +27,9 @@
         "prefer-promise-reject-errors": ["error", { "allowEmptyReject": true }],
         "react/jsx-indent": ["error", 4],
         "semi": ["error", "always", { "omitLastInOneLineBlock": true }],
+
+        "react-hooks/rules-of-hooks": "error",
+        "react-hooks/exhaustive-deps": "error",
 
         "camelcase": "off",
         "comma-dangle": "off",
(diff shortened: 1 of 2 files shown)


martinpitt:
Actually, let's use the corresponding eslint plugin for that. With a (lame) diff like this:

```diff
--- src/app.jsx
+++ src/app.jsx
@@ -26,7 +26,13 @@ const _ = cockpit.gettext;
 export function Application() {
     const [hostname, setHostname] = useState(_("Unknown"));
 
+    function blah() {
+        const [a, b] = useState(0);
+        console.log(a, b);
+    }
+
     cockpit.file('/etc/hostname').watch(content => setHostname(content.trim()));
+    blah();
 
     return (
         <div className="container-fluid">

```

it will then correctly complain:
```
/var/home/martin/upstream/starter-kit/src/app.jsx
  30:24  error  React Hook "useState" is called in function "blah" which is neither a React function component or a custom React Hook function  react-hooks/rules-of-hooks
```

mvollmer:
Hooks are cool, but also quite magical.  But as long as eslint prevents mistakes, we should be good.

martinpitt:
Hang on, this is bogus. We don't want to run the `cockpit.file().watch()` on every render. So having this as a functional component isn't right by design. starter-kit is just too simple for actually using hooks.

However, I think adding the eslint rules is still beneficial for users who actually clone and use that. I rewrote the PR accordingly.

mvollmer:
> We don't want to run the `cockpit.file().watch()` on every render.

Ouch, indeed.  Would effect hooks work for this?

PR #290, opened 2020-03-10
Various test/build system fixes

martinpitt:


martinpitt:
@marusak, @skobyda  : These changes may be interesting for c-podman and c-certificates as well. In particular, it makes iterating with `make check` a lot nicer.

PR #291, opened 2020-03-10
test: Use test scenario for selecting browser

martinpitt:
Similar to what cockpit-composer does. This is rather simplistic, but
does fine until we need more complex scenarios.

Also use plain shell in test/run, there are no bashisms.

### DIFF
--- a/test/run
+++ b/test/run
@@ -1,4 +1,8 @@
-#! /bin/bash
+#! /bin/sh
+set -eu
+
 # This is the expected entry point for Cockpit CI; will be called without
-# arguments but with an appropriate $TEST_OS
+# arguments but with an appropriate $TEST_OS, and optionally $TEST_SCENARIO
+
+[ -z "${TEST_SCENARIO:-}" ] || export TEST_BROWSER="$TEST_SCENARIO"
 make check


martinpitt:
Needs https://github.com/cockpit-project/bots/pull/627 , then I'll trigger the firefox test.

martinpitt:
Interesting! It [failed](https://logs.cockpit-project.org/logs/pull-291-20200310-111030-a476c9ee-fedora-32-firefox/log.html), but it works locally in a cockpit/tasks container (through podman). That's kind of what I was hoping for, as the same failure affects cockpit. But with starter-kit it's much faster to investigate.

martinpitt:
I. e. I think this PR by itself is fine, the scenario is obviously respected and applied. But of course there's no urgency, and we can leave this open to investigate the CI failure.

martinpitt:
I'm trying this in a local CentOS 7 VM (which should be fairly close to what's running on CentOS CI, and our e2e machines run RHEL 7), and at least with a single test run it seems to be reliable:
```
./vm-run -C2 -M4096 services  # or centos-7; I just happened to have services downloaded already
# free some space
docker rmi selenium/node-{firefox,chrome}-debug:3 selenium/hub:3

mkdir -p /cache/images/
# now scp fedora-31 image from host to that dir
chown -R 1111:1111 /cache/

chmod 666 /dev/kvm
docker network create --driver bridge ndbg
docker run -it --name=ndbg --device=/dev/kvm --rm --volume=/cache:/cache:Z --shm-size=1024m --network=ndbg cockpit/tasks /bin/bash
```
and inside the container:
```
git clone -b firefox --depth=1 https://github.com/martinpitt/starter-kit.git
cd starter-kit
TEST_AUDIT_NO_SELINUX=1 TEST_OS=fedora-31 TEST_BROWSER=firefox make check
```
This mirrors pretty much exactly what's happening in our CI. I tried (almost) the same commands on cockpit-7 (that has oci-kvm-hook installed, and a different cache location):
```
docker run -it --name=ndbg --rm --volume=/var/cache/cockpit-tasks:/cache:rw --shm-size=1024m --network=ndbg cockpit/tasks /bin/bash
```
and in that container it works fine as well. I also tried to create two (network and container `ndbg2`) and run the test in parallel, which also works.

So no luck yet, but I want to put the commands somewhere.

martinpitt:
I had a closer look -- this morning, https://github.com/cockpit-project/cockpit/pull/13670 [succeeded](https://logs.cockpit-project.org/logs/pull-13659-20200310-074956-bf252a25-fedora-31-firefox/log) on cockpit-7. But the [run here](https://logs.cockpit-project.org/logs/pull-291-20200310-111030-a476c9ee-fedora-32-firefox/log.html) or a [different failure for cockpit](https://logs.cockpit-project.org/logs/pull-13673-20200309-022355-ac617f9c-fedora-31-firefox/log) failed on cockpit-10. With the above reproducer I can reliably reproduce the crash, so they are useful after all. :-)

Going through some other logs: [ci-srv-01 pass](https://logs.cockpit-project.org/logs/pull-13682-20200309-102640-00f81d28-fedora-31-firefox/log), [another ci-srv-01 pass](https://logs.cockpit-project.org/logs/pull-13693-20200310-103748-05c609b1-fedora-31-firefox/log) [ci-srv-06 fail](https://logs.cockpit-project.org/logs/pull-13694-20200310-105913-1c06291a-fedora-31-firefox/log), [centos-tasks-xjfc9 fail](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-13692-20200310-104033-52c6bf9e-fedora-31-firefox/log) (but that test has failed on CentOS for a long time already), [ci-srv-04 fail](https://logs.cockpit-project.org/logs/pull-13690-20200310-084046-97a4a906-fedora-31-firefox/log)

As we have more than one affected machine, I'm going to do some experiments on cockpit-10.

First, I stopped all regular tasks containers, so that the test runs on its own, but kept my ndbg container. Now it's even worse -- it gets this far and then hangs forever:

```
+ systemctl start cockpit.socket
-> switch to frame None
-> Page.navigate({"url": "http://127.0.0.2:9091/starter-kit"})
*** You are running in headless mode.
Started /usr/bin/firefox (pid 1929) on port 9924
*** You are running in headless mode.
out of memory: 0x0000000000000038 bytes requested
out of memory: 0x0000000000000070 bytes requested
[Child 2033, Chrome_ChildThread] WARNING: pipe error (3): Connection reset by peer: file /builddir/build/BUILD/firefox-69.0.3/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 358
[Child 2004, Chrome_ChildThread] WARNING: pipe error (3): Connection reset by peer: file /builddir/build/BUILD/firefox-69.0.3/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 358
Exiting due to channel error.
Exiting due to channel error.
```
I cancelled it after > 20 minutes.

Next I restarted ndbg, so that it was a clean container start as the only docker container on the machine. Still the same "indefinite hang". In top, firefox just goes crazy, 300% CPU for `firefox` and another 150% for `WebExtensions`, and firefox keeps piling up memory (13.6 GB after 3 minutes). That is a good reminder that we should limit the memory of our containers. Our kubernetes resources limit a container to 16 GiB, I'll add that to our pittishift as well.

Next step, I stopped and restarted docker.service. I get a tiny bit further (`wait: ph_is_present("#login")`, and then that times out), but otherwise firefox is still going mad.

Finally, big hammer: I rebooted cockpit-10, and only ran ndbg (no cockpit-tasks@*.service yet), and this finally works (not totally surprising). So there is some kind of cruft building up over time in the host's kernel that seems hard to eliminate/reset.

Another observation: when it works (locally and on cockpit-10), I also get this warning:

    CDP: {"source":"javascript","level":"warning","text":"Password fields present on an insecure (http://) page. This is a security risk that allows user login credentials to be stolen.","timestamp":1583852273212,"url":"http://127.0.0.2:9091/starter-kit","lineNumber":0}

so this is clearly unrelated to the failure.

martinpitt:
So barring infra problems, this generally works now, and I believe we can land this. Note that the test isn't enabled by default in CI yet, but I think we should do that.

PR #315, opened 2020-05-19
main: Stop importing patternfly.css

KKoukiou:
Instead install and import the styles from PF4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 NODE_MODULES_TEST=package-lock.json
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/index.html
+SASS=$(CURDIR)/node_modules/.bin/sass
 
 all: $(WEBPACK_TEST)
 
@@ -65,7 +66,7 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
 %.spec: %.spec.in
 	sed -e 's/%{VERSION}/$(VERSION)/g' $< > $@
 
-$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js $(patsubst %,dist/po.%.js,$(LINGUAS))
+$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js dist/patternfly-4-cockpit.css $(patsubst %,dist/po.%.js,$(LINGUAS))
 	NODE_ENV=$(NODE_ENV) npm run build
 
 watch:
@@ -75,9 +76,11 @@ clean:
 	rm -rf dist/
 	[ ! -e cockpit-$(PACKAGE_NAME).spec.in ] || rm -f cockpit-$(PACKAGE_NAME).spec
 
+# patternfly-4-cockpit.css needs to be shipped in the release tarball to avoid rebuilding the webpack, but we don't need to install it
 install: $(WEBPACK_TEST)
 	mkdir -p $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)
 	cp -r dist/* $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)
+	rm $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)/patternfly-4-cockpit.css
 	mkdir -p $(DESTDIR)/usr/share/metainfo/
 	cp org.cockpit-project.$(PACKAGE_NAME).metainfo.xml $(DESTDIR)/usr/share/metainfo/
 
@@ -155,6 +158,10 @@ test/common:
 	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 219
 	git checkout --force FETCH_HEAD -- test/common
 	git reset test/common
+
+dist/patternfly-4-cockpit.css: src/lib/patternfly-4-cockpit.scss $(NODE_MODULES_TEST)
+	$(SASS) $< $@ --source-map --source-map-urls absolute --style compressed --load-path node_modules && \
+	sed -i -f src/lib/patternfly.sed $@
 
 $(NODE_MODULES_TEST): package.json
 	# if it exists already, npm install won't update it; force that so that we always get up-to-date packages
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,6 @@ VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 NODE_MODULES_TEST=package-lock.json
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/index.html
-SASS=$(CURDIR)/node_modules/.bin/sass
 
 all: $(WEBPACK_TEST)
 
@@ -66,7 +65,7 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
 %.spec: %.spec.in
 	sed -e 's/%{VERSION}/$(VERSION)/g' $< > $@
 
-$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js dist/patternfly-4-cockpit.css $(patsubst %,dist/po.%.js,$(LINGUAS))
+$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js $(patsubst %,dist/po.%.js,$(LINGUAS))
 	NODE_ENV=$(NODE_ENV) npm run build
 
 watch:
@@ -76,11 +75,9 @@ clean:
 	rm -rf dist/
 	[ ! -e cockpit-$(PACKAGE_NAME).spec.in ] || rm -f cockpit-$(PACKAGE_NAME).spec
 
-# patternfly-4-cockpit.css needs to be shipped in the release tarball to avoid rebuilding the webpack, but we don't need to install it
 install: $(WEBPACK_TEST)
 	mkdir -p $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)
 	cp -r dist/* $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)
-	rm $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)/patternfly-4-cockpit.css
 	mkdir -p $(DESTDIR)/usr/share/metainfo/
 	cp org.cockpit-project.$(PACKAGE_NAME).metainfo.xml $(DESTDIR)/usr/share/metainfo/
 
@@ -158,10 +155,6 @@ test/common:
 	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 219
 	git checkout --force FETCH_HEAD -- test/common
 	git reset test/common
-
-dist/patternfly-4-cockpit.css: src/lib/patternfly-4-cockpit.scss $(NODE_MODULES_TEST)
-	$(SASS) $< $@ --source-map --source-map-urls absolute --style compressed --load-path node_modules && \
-	sed -i -f src/lib/patternfly.sed $@
 
 $(NODE_MODULES_TEST): package.json
 	# if it exists already, npm install won't update it; force that so that we always get up-to-date packages
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -19,6 +19,7 @@
 
 import cockpit from 'cockpit';
 import React from 'react';
+import { Alert, Card, CardHead, CardHeader, CardHeadMain, Title } from '@patternfly/react-core';
 import './app.scss';
 
 const _ = cockpit.gettext;
@@ -35,12 +36,19 @@ export class Application extends React.Component {
 
     render() {
         return (
-            <div className="container-fluid">
-                <h2>Starter Kit</h2>
-                <p>
-                    { cockpit.format(_("Running on $0"), this.state.hostname) }
-                </p>
-            </div>
+            <Card>
+                <CardHead>
+                    <CardHeadMain>
+                        <Title headingLevel="h2" size="3xl">Starter Kit</Title>
+                    </CardHeadMain>
+                </CardHead>
+                <CardHeader>
+                    <Alert
+                        variant="default"
+                        title={ cockpit.format(_("Running on $0"), this.state.hostname) }
+                    />
+                </CardHeader>
+            </Card>
         );
     }
 }
--- a/src/index.html
+++ b/src/index.html
@@ -22,7 +22,6 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <link rel="stylesheet" href="../base1/patternfly.css">
     <link rel="stylesheet" href="index.css">
 
     <script type="text/javascript" src="../base1/cockpit.js"></script>
@@ -30,7 +29,7 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
     <script type="text/javascript" src="index.js"></script>
 </head>
 
-<body>
+<body class="pf-m-redhat-font">
     <div id="app"></div>
 </body>
 </html>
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -156,6 +157,12 @@ test/common:
 	git checkout --force FETCH_HEAD -- test/common
 	git reset test/common
 
+SASS = ./node_modules/.bin/sass
```
martinpitt, line 160: Let's put that at the top

Makefile
```diff
@@ -156,6 +157,12 @@ test/common:
...
+build-patternfly:
```
martinpitt, line 161: This is a phony rule (i. e. not an actual file), and thus will break make dependencies. I. e. this will now run on each invocation of `make`, which will in turn require re-building the rpm and the VM (which is expensive), even if nothing changed. If you type `make` twice, the second one should be a no-op.

Let's instead make this a rule that specifies the actual inputs and outputs. Then you can also use the pre-defined make vars:

```make
dist/patternfly-4-base.css: $(srcdir)/lib/patternfly-4-cockpit.scss
	$(SASS) $< $@ --source-map ...
```

src/app.jsx
```diff
@@ -35,12 +36,19 @@ export class Application extends React.Component {
...
-            <div className="container-fluid">
-                <h2>Starter Kit</h2>
-                <p>
-                    { cockpit.format(_("Running on $0"), this.state.hostname) }
-                </p>
-            </div>
+            <Card>
+                <CardHead>
+                    <CardHeadMain>
+                        <Title headingLevel="h2" size="3xl"> Starter Kit </Title>
```
martinpitt, line 42: Please drop the wrapping whitespace inside the Title.

src/app.scss
```diff
@@ -1,3 +1,9 @@
+/*
+ * PF4 overrides need to come after the JSX components imports because
+ * these are importing CSS stylesheets that we are trying to override
```
martinpitt, line 3: s/trying to override/overriding/

But this comment looks a bit weird here -- this file doesn't import any PF4 stylesheets. Is it possible to import app.scss in src/index.js instead? If not, let's move the comment to index.html in between importing pf-4-base.css and index.css? (although that's rather obvious even without the comment)

src/lib/_fonts.scss
```diff
@@ -0,0 +1,32 @@
+@mixin printRedHatFont(
```
martinpitt, line 1: Why "_fonts", and not just "fonts"?

src/lib/patternfly-4-cockpit.scss
```diff
@@ -0,0 +1,14 @@
...
+$pf-global--font-path: '/static/fonts';
+$pf-global--fonticon-path: '/base1/fonts';
+$pf-global--disable-fontawesome: true !default; // Disable Font Awesome 5 Free
+@import "../../node_modules/@patternfly/patternfly/sass-utilities/all";
+@import "../../node_modules/@patternfly/patternfly/variables";
+@import "../../node_modules/@patternfly/patternfly/base";
+@import "../../node_modules/@patternfly/patternfly/themes";
+@import "../../node_modules/@patternfly/patternfly/icons";
+
+@import "./fonts";
```
martinpitt, line 14: That looks inconsistent -- file is "_fonts.scss", and it tries to import "fonts"?

src/lib/patternfly-4-cockpit.scss
```diff
@@ -0,0 +1,14 @@
+/* Here we need to import patternfly-base.scss but the fonts file from PF4 is containing
+ * urls to font formats we are not interested in including.
+ * @import '@patternfly/patternfly/patternfly-base.scss';
```
martinpitt, line 3: That comment doesn't really explain the conclusion -- either this import is the desired form, and the variable settings and individual imports below are a workaround, then it should point that out. Or this generally isn't meant to work, then let's drop the commented-out `@import` completely.

I. e. the code below makes it look like you could first define the three variables and then import patternfly-base.scss?

src/lib/patternfly.sed
```diff
@@ -0,0 +1,2 @@
+s/src: url(.*eot[^'"]*['"]);$//
```
martinpitt, line 1: Let's add a comment to that file why we need it, and where the original source is (cockpit's pkg/lib/)

src/lib/patternfly-4-cockpit.scss
```diff
@@ -0,0 +1,14 @@
+/* Here we need to import patternfly-base.scss but the fonts file from PF4 is containing
```
martinpitt, line 1: Let's add a comment where the original source of that file is (or rather, "is going to be :grin: ), and that it should be kept in sync with cockpit. Same for fonts.scss and any other synced file.

src/app.scss
KKoukiou: patternfly-4-overrides is not overriding rules from patternfly-4-base.css :) It's fixing the stylesheets that are imported when we include React components form `@patternfly/react-core`.
patternfly-4-base.css is just the base so that the per-components stylesheets can be cleanly applied - imagine is as a CSS reset.

So, these fixes (patternfly-4-overrides.scss) have to be after all the PF4 components CSS is in the resulting dist/index.css produced file.
`mini-css-extract-plugin` does the work for us, since it takes all the CSS that was included in the dist/index.js files from the css/scss loaders and puts it in the dist/index.css in the order that these files where imported.
So after the scss/css loader run we do have a dist/index.js file full of CSS rules, in the order of PF4 components imports, app.scss compiled contents. Which is exactly what we want. 

However I understand the confusion - it's rather not clear - thus I moved the overrides in the index.js file but kept the app.scss where it was not make it clear that these are not connected.

src/lib/_fonts.scss
KKoukiou: A sass file starting with an underscore is a partial. It is a good way to keep your styles separated into logical sections. These files all get merged on compilation when you use @import.
https://sass-lang.com/guide

When you import a _partial.scss file like `import 'partial` an inline import takes place.
In this case there was not need to do that since I am not using variables from outside this file, but I think it's a good practice.

src/lib/patternfly-4-cockpit.scss
KKoukiou: See above comment.

src/lib/_fonts.scss
martinpitt: Agreed. Thanks for the explanation!

src/lib/patternfly-4-cockpit.scss
KKoukiou: `@import '@patternfly/patternfly/patternfly-base.scss'` is what PF4 exposes for us normally.
It's consisting from all the imports I do, plus one more that I replaced with the `@import fonts` line.
The line that is missing is:
`@import "../../node_modules/@patternfly/patternfly/fonts";`

Makefile
```diff
@@ -156,6 +157,12 @@ test/common:
...
+	F=$(srcdir)/lib/patternfly-4-cockpit.scss && \
+	$(SASS) $$F dist/patternfly-4-base.css --source-map --load-path node_modules && \
+	sed -i -f $(srcdir)/lib/patternfly.sed dist/patternfly-4-base.css
```
martinpitt, line 164: Please let SASS output to $@.tmp, mangle that, and as the very last step `mv $@.tmp $@`. In the case of errors this will make sure that we don't have half-written/broken result files.

src/lib/patternfly-4-cockpit.scss
KKoukiou: I added the commnents and links but of course these won't be available until cockpit is adjusted to this idea as well.

src/lib/_fonts.scss
martinpitt, line 1: Let's please also add a comment where to sync this from.

src/lib/patternfly-4-overrides.scss
```diff
@@ -0,0 +1,39 @@
+/*** PF4 overrides ***/
```
martinpitt, line 1: Same here, let's add a comment where the original file is (... will be soon)

Makefile
```diff
@@ -156,6 +158,11 @@ test/common:
 	git checkout --force FETCH_HEAD -- test/common
 	git reset test/common
 
+dist/patternfly-4-base.css: $(srcdir)/lib/patternfly-4-cockpit.scss
+	$(SASS) $< $@.tmp --source-map --source-map-urls absolute --style compressed --load-path node_modules && \
```
martinpitt, line 162: When I run `make` in a clean tree, it fails with
```
/bin/sh: ./node_modules/.bin/sass: No such file or directory
```
 Please add `$(WEBPACK_TEST)` as another (second!) dependency to that rule, to make sure that this runs after npm install.

Also, this somehow needs to specify the name of the .map file -- right now it builds `patternfly-4-base.css.tmp.map`. If that's not possible, then let's drop the `.tmp` again (but that would reintroduce the race condition again)

src/lib/patternfly-4-cockpit.scss
```diff
@@ -0,0 +1,14 @@
...
+
+/* Set fake font and icon path variables - we are going to indentify these through
+ * patternfly.sed and remove the relevant font-face declarations
+ */
+$pf-global--font-path: 'patternfly-fonts-fake-path';
+$pf-global--fonticon-path: 'patternfly-icons-fake-path';
+$pf-global--disable-fontawesome: true !default; // Disable Font Awesome 5 Free
+@import '@patternfly/patternfly/patternfly-base.scss';
+
+/* Import out own fonts since the PF4 font-face rules are filtered out with patternfly.sed */
```
martinpitt, line 13: s/out/our/

Makefile
KKoukiou: It's not possible to specify the source map name. It's rare that we make changed to all these PF4 files so lets not bother so much about it, what do you think?

martinpitt: Ack, then let's drop the .tmp stuff again.

martinpitt:
Let's clarify the commit message on squashing, or if it needs another force-push:

    Stop importing cockpit's base1/patternfly.css

    This is deprecated API and will be dropped at some point, in favor
    of projects shipping their own CSS.

    Install and import the styles from PF4 now.

martinpitt:
Still broken CSS with `TEST_OS=fedora-31 make vm && bots/vm-run -s cockpit.socket fedora-31`

src/index.html
```diff
@@ -22,15 +22,15 @@
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <link rel="stylesheet" href="../base1/patternfly.css">
+    <link rel="stylesheet" href="patternfly-4-base.css">
```
martinpitt, line 25: Still broken, I'm afraid :crying_cat_face: But my dist/ folder doesn't actually have that file:
```
❱❱❱ ls -l dist/
total 1248
-rw-rw-r--. 1 martin martin   19071 27. Mai 16:03 index.css
-rw-rw-r--. 1 martin martin   27453 27. Mai 16:03 index.css.map
-rw-rw-r--. 1 martin martin    1284 27. Mai 16:03 index.html
-rw-rw-r--. 1 martin martin   96367 27. Mai 16:03 index.min.js.gz
-rw-rw-r--. 1 martin martin 1114220 27. Mai 16:03 index.min.js.map
-rw-rw-r--. 1 martin martin     162 27. Mai 16:03 manifest.json
-rw-rw-r--. 1 martin martin     512 27. Mai 16:03 po.de.js
```

Is it supposed to be a disted file, or does that get built into the webpack?

martinpitt:
Fixed the makefile, now it works fine:
```diff
--- Makefile
+++ Makefile
@@ -15,7 +15,7 @@ WEBPACK_TEST=dist/index.html
 SASS = ./node_modules/.bin/sass
 srcdir=./src
 
-all: $(WEBPACK_TEST) dist/patternfly-4-base.css
+all: $(WEBPACK_TEST)
 
 #
 # i18n
@@ -67,7 +67,7 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
 %.spec: %.spec.in
        sed -e 's/%{VERSION}/$(VERSION)/g' $< > $@
 
-$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js $(patsubst %,dist/po.%.js,$(LINGUAS))
+$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js dist/patternfly-4-base.css $(patsubst %,dist/po.%.js,$(LINGUAS))
        NODE_ENV=$(NODE_ENV) npm run build
 
 watch:
@@ -158,7 +158,7 @@ test/common:
        git checkout --force FETCH_HEAD -- test/common
        git reset test/common
 
-dist/patternfly-4-base.css: $(srcdir)/lib/patternfly-4-cockpit.scss $(WEBPACK_TEST)
+dist/patternfly-4-base.css: $(srcdir)/lib/patternfly-4-cockpit.scss $(NODE_MODULES_TEST)
        $(SASS) $< $@ --source-map --source-map-urls absolute --style compressed --load-path node_modules && \
        sed -i -f $(srcdir)/lib/patternfly.sed $@
 
```

KKoukiou:
> Fixed the makefile, now it works fine:
> 
> ```diff
> --- Makefile
> +++ Makefile
> @@ -15,7 +15,7 @@ WEBPACK_TEST=dist/index.html
>  SASS = ./node_modules/.bin/sass
>  srcdir=./src
>  
> -all: $(WEBPACK_TEST) dist/patternfly-4-base.css
> +all: $(WEBPACK_TEST)
>  
>  #
>  # i18n
> @@ -67,7 +67,7 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
>  %.spec: %.spec.in
>         sed -e 's/%{VERSION}/$(VERSION)/g' $< > $@
>  
> -$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js $(patsubst %,dist/po.%.js,$(LINGUAS))
> +$(WEBPACK_TEST): $(NODE_MODULES_TEST) $(shell find src/ -type f) package.json webpack.config.js dist/patternfly-4-base.css $(patsubst %,dist/po.%.js,$(LINGUAS))
>         NODE_ENV=$(NODE_ENV) npm run build
>  
>  watch:
> @@ -158,7 +158,7 @@ test/common:
>         git checkout --force FETCH_HEAD -- test/common
>         git reset test/common
>  
> -dist/patternfly-4-base.css: $(srcdir)/lib/patternfly-4-cockpit.scss $(WEBPACK_TEST)
> +dist/patternfly-4-base.css: $(srcdir)/lib/patternfly-4-cockpit.scss $(NODE_MODULES_TEST)
>         $(SASS) $< $@ --source-map --source-map-urls absolute --style compressed --load-path node_modules && \
>         sed -i -f $(srcdir)/lib/patternfly.sed $@
>  
> ```

now with this it fails actualy with /bin/sh: ./node_modules/.bin/sass: No such file or directory

I run git clean -fdx first - so it removed node_modules.

martinpitt:
Ah, that didn't work, we have to ship the generated .css. Otherwise the rpm build will try to rebuild it again (as Make rules depend on it), and that doesn't work because node_modules/ (and thus sass and webpack) are not available in the rpm build environment. Reverted back to shipping the .css in dist/, and just not installing into the rpm (it would be harmless, but unnecessary).

I tested this with a local build from scratch, and a VM+RPM build, and it's working fine now. RPM content looks reasonable, too.

Sorry, that took a while :-)

martinpitt:
[travis failure](https://travis-ci.org/github/cockpit-project/starter-kit/builds/691834011) as it doesn't run `make`, but `npm run build`. Switched to `make` for now, maybe we find a way to build the sass → css from within webpack?

martinpitt:
I think we need to fix README at the very least, which documents running `npm run build`. This isn't working any more. And neither is `make watch`, as it's missing the dependencies. Perhaps we should move the css build and seddery to package.json?

martinpitt:
I was wondering if we can do this within webpack, to avoid breaking the `npm run build/watch` toolchain, and avoid all these build system hacks. I took some time to more properly understand what webpack loaders are all about, and found [string-replace-loader](https://github.com/Va1/string-replace-loader) which is essentially sed for webpack.

I now have this at the point where it works. I diffed the resulting css, and it's (by and large, modulo some whitespace) identical. I verified correct fonts (RedHatText, and loaded from /cockpit/static/) in both running from local tree as well as a clean build into rpm.

I added that as a separate commit, so that it's easier to discuss. If you like it, we can just squash it, otherwise take it out again. @KKoukiou , WDYT?

PR #317, opened 2020-05-22
Bump up version of test/common

marusak:
There has been a lot of improvements since 213 and also due to new
privilege escalation this is need.


marusak:
Argh, `bash: /usr/share/cockpit/packagekit/override.json: No such file or directory` - this was not the smartest move, I guess we need there ` || true`.

marusak:
Adding quick hack so we don't need to wait until it is fixed, new version released and we can bump up the tag.

PR #321, opened 2020-05-28
Drop Vagrantfile

martinpitt:
Nobody in the development team has ever really used that, it needs a ton
of dependencies, and doesn't work in toolbox/Silverblue. Our own
vm-run/image-customize etc. machinery is efficient enough, and proven to
work.


martinpitt:
@larskarlitski : I think you are the one who still has the closest affinity to vagrant. Do you mind reviewing/commenting?

PR #325, opened 2020-06-02
test: Adjust to new navigation in Cockpit 220

martinpitt:
The parent element `#navbar-dropdown` exists in both old and new
versions.


martinpitt:
Should fix [this failure](https://logs.cockpit-project.org/logs/pull-926-20200602-060123-a945dedc-fedora-32-cockpit-project-starter-kit/log.html)

PR #333, opened 2020-06-24
Stop importing cockpit's base1/patternfly.css

KKoukiou:


KKoukiou:
Sorry that was meant for session-recording - but because it has the same source fork it was automatically created here.

PR #345, opened 2020-07-23
Exclude node_modules from webpack watching

martinpitt:
This recently started to overflow inotify, and we don't need that --
it's fine to restart webpack when changing package.json, and it happens
rarely enough.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -118,6 +118,9 @@ module.exports = {
     resolve: {
         modules: [ nodedir ],
     },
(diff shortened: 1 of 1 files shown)


martinpitt:
See https://webpack.js.org/configuration/watch/ . Inspired by https://github.com/cockpit-project/cockpit/pull/14396

martinpitt:
Once that lands, I'll cherry-pick it into cockpit-podman as well.

PR #357, opened 2020-08-24
Copy patternfly import files from cockpit project automatically

KKoukiou:
Similarly to bots and test/common copying technique.

This should be blocked till we release 227 cockpit.


martinpitt:
I like the idea! As a bit of bikeshedding, would it be ok to keep them in src/lib/patternfly/?

Makefile
```diff
@@ -157,6 +157,12 @@ test/common:
 	git checkout --force FETCH_HEAD -- test/common
 	git reset test/common
 
+patternfly:
+	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 227
```
martinpitt, line 161: If you want to test this early, you can also fetch the origin without depth and check out a SHA.

KKoukiou:
> I like the idea! As a bit of bikeshedding, would it be ok to keep them in src/lib/patternfly/?

Sure, I will need to repost because we don't fetch these from base1 any more.

martinpitt:
error: pathspec 'pkg/lib/patternfly' did not match any file(s) known to git

KKoukiou:
> ```
> git checkout --force FETCH_HEAD -- pkg/lib/patternfly
> error: pathspec 'pkg/lib/patternfly' did not match any file(s) known to git
> ```

!!  it worked locally from a completely clean git checkout, and also the directory clearly exists https://github.com/cockpit-project/cockpit/tree/227/pkg/lib/patternfly :confused:

martinpitt:
The [travis failure](https://travis-ci.org/github/cockpit-project/starter-kit/builds/724132626) is simple to explain -- .travis.yml just runs `npm install; npm run build`, so the git cloning never comes into play. I pushed a fix.

martinpitt:
I can reproduce the failure locally in a bots checkout, using the exact command from the log. The reason is quite funny:
```
$ ./make-checkout --verbose --repo=cockpit-project/starter-kit --rebase=master pull/357/head && cd make-checkout-workdir && TEST_OS=fedora-32 COCKPIT_BOTS_REF=master ../tests-invoke --pull-number 357 --revision a6f650a8535bb6789fcda61564f4fdac05e0d110 --repo cockpit-project/starter-kit
[...]
fatal: Unable to create '/var/home/martin/upstream/bots/make-checkout-workdir/.git/shallow.lock': File exists. fetche
Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
```

Indeed bots, test/common, and src/lib/patternfly all call git, and you can't call more than one instance in parallel without additional help. test/run calls `make check` directly, and as test/common and bots are independent targets, make tries to run them in parallel.

A simple `make check` from a clean checkout reproduces the same thing.

I'm not sure if that's really the same root cause as in CI, the log looks quite different. But we need to fix that either way.

martinpitt:
I fixed the "parallel git" issue, and can now reproduce the same failure as in CI with a simple `make check`.

```
+ /usr/bin/make install DESTDIR=/var/home/martin/upstream/starter-kit/build/cockpit-starter-kit-1-1.fc33.x86_64 'INSTALL=/usr/bin/install -p'
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[1]: Entering directory '/var/home/martin/upstream/starter-kit/rpmbuild/cockpit-starter-kit'
error: Unable to open cockpit-starter-kit.spec.in: No such file or directory
error: query of specfile cockpit-starter-kit.spec.in failed, can't parse
error: Unable to open cockpit-starter-kit.spec.in: No such file or directory
error: query of specfile cockpit-starter-kit.spec.in failed, can't parse
error: Unable to open cockpit-starter-kit.spec.in: No such file or directory
error: query of specfile cockpit-starter-kit.spec.in failed, can't parse
git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 227
remote: Enumerating objects: 1514, done.
remote: Counting objects: 100% (1514/1514), done.
remote: Compressing objects: 100% (1389/1389), done.
remote: Total 1500 (delta 253), reused 434 (delta 18), pack-reused 0
Receiving objects: 100% (1500/1500), 3.96 MiB | 4.66 MiB/s, done.
Resolving deltas: 100% (253/253), completed with 4 local objects.
From https://github.com/cockpit-project/cockpit
 * tag                 227        -> FETCH_HEAD
mkdir -p pkg/lib/patternfly && git add pkg/lib/patternfly
git checkout --force FETCH_HEAD -- pkg/lib/patternfly
error: pathspec 'pkg/lib/patternfly' did not match any file(s) known to git
make[1]: *** [Makefile:163: src/lib/patternfly] Error 1
make[1]: Leaving directory '/var/home/martin/upstream/starter-kit/rpmbuild/cockpit-starter-kit'
error: Bad exit status from /var/tmp/rpm-tmp.IUbdR9 (%install)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.IUbdR9 (%install)
make: *** [Makefile:117: cockpit-starter-kit-1-1.fc33.noarch.rpm] Error 1
```

This happens during `make vm`, as it tries to rebuild the webpack as part of the RPM build inside the VM. This must not happen, the release tarball must contain all source files *and* the ready-made webpack. For once, the release tarball does not contain these files, and it's probably also not a good idea to have a make dependency on a directory instead of a file.

martinpitt:
Fixed dist-git tarballs and depencies, indeed the dir dependency broke time stamps as well, due to the tarball not storing directories:

```
    + /usr/bin/tar -xvvof -
[...]
drwxr-xr-x                  Creating directory: cockpit-starter-kit/src/lib
drwxr-xr-x                  Creating directory: cockpit-starter-kit/src/lib/patternfly
-rw-r--r-- martin/martin  1398 2020-09-07 14:50 cockpit-starter-kit/src/lib/patternfly/_fonts.scss
-rw-r--r-- martin/martin   510 2020-09-07 14:50 cockpit-starter-kit/src/lib/patternfly/patternfly-4-cockpit.scss
-rw-r--r-- martin/martin   963 2020-09-07 14:50 cockpit-starter-kit/src/lib/patternfly/patternfly-4-overrides.scss
```
i. e. the timestamp of src/lib/patternfly would always be newer than the webpack.

martinpitt:
I'm fine with this now, but as I changed it heavily, I'd like someone else to do the final review.

martinpitt (approved):

PR #371, opened 2020-09-21
Makefile: Serialize git commands using flock

martinpitt:
We previously made directories depend on files for the synthetic
serialization of git commands. git does not track directories, and
time stamps across repos (starter-kit vs. bots) work rather poorly --
for example, this broke "bots" being a symlink to a previously existing
checkout.

Use file locking instead, which is a more direct way of ensuring that
there is only one git operation that affects the current checkout. We
don't need to flock bots/, as that's an independent git repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,20 +153,18 @@ bots:
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
-	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 228
-	git checkout --force FETCH_HEAD -- test/common
-	git reset test/common
+	flock Makefile sh -ec '\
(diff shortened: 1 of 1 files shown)


martinpitt:
Before, in a clean clone:
```
$ ln -s ../bots
$ make bots
git clone --quiet --reference-if-able ${XDG_CACHE_HOME:-$HOME/.cache}/cockpit-project/bots https://github.com/cockpit-project/bots.git
fatal: destination path 'bots' already exists and is not an empty directory.
make: *** [Makefile:149: bots] Error 128
``

Now it's back to before commit cfced1900069ed:
```
$ make bots
make: 'bots' is up to date.
```

marusak:
Thanks, seems reasonable. Can you please rebase though as there is a conflict? Thanks!

martinpitt:
Rebased.

marusak (approved):
Thanks!

PR #373, opened 2020-09-28
npm: Upgrade development dependencies

martinpitt:
compression-webpack-plugin fixes https://www.npmjs.com/advisories/1548

Adjust to new copy-webpack-plugin API.


martinpitt:
This [build failure](https://logs.cockpit-project.org/logs/pull-373-20200928-051309-4b32454d-fedora-32/log.html) happens during `make rpm`, for some reason it wants to rebuild the webpack.

There is a difference in the generated tarball:
```
-cockpit-starter-kit/dist/index.html
+cockpit-starter-kit/dist/index.html.gz
```
which influences/breaks `WEBPACK_TEST`. But I can't change it to `index.html.gz` as that breaks NODE_ENV=devel. So I moved it to index.css instead, which works in both modes.

martinpitt:
The [travis failure](https://travis-ci.org/github/cockpit-project/starter-kit/builds/730848463?utm_source=github_status&utm_medium=notification) looks odd as well:
```
home/travis/build/cockpit-project/starter-kit/node_modules/copy-webpack-plugin/node_modules/p-limit/index.js:30

		} catch {}
		        ^
SyntaxError: Unexpected token {
```
Perhaps it's because we are still specifying an ancient `node_js: 8` in our .travis.yml? It's not even supported any more in https://docs.travis-ci.com/user/languages/javascript-with-nodejs/

marusak (approved):

PR #375, opened 2020-09-29
Use standard "translate" marker in HTML, fix PO template build

martinpitt:
Cockpit's test-static-code complains about `translatable`, so let's use
the correct attribute to avoid spreading it further.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,7 +17,7 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
 -->
 <html lang="en">
 <head>
-    <title translatable="yes">Cockpit Starter Kit</title>
+    <title translate>Cockpit Starter Kit</title>
     <meta charset="utf-8">
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,4 @@ node_js:
   - "lts/*"
 script:
   - make
+  - make po/starter-kit.pot
--- a/po/html2po
+++ b/po/html2po
@@ -45,9 +45,9 @@ try {
 }
 
 var opts = stdio.getopt({
-    directory: { key: "d", args: 1, description: "Base directory for input files" },
+    directory: { key: "d", args: 1, description: "Base directory for input files", default: "." },
     output: { key: "o", args: 1, description: "Output file" },
-    from: { key: "f", args: 1, description: "File containing list of input files" },
+    from: { key: "f", args: 1, description: "File containing list of input files", default: "" },
 });
 
 if (!opts.from && opts.args.length < 1) {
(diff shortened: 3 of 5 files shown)


martinpitt:
Note that our integration tests check that the title actually gets translated.

src/index.html
```diff
@@ -17,7 +17,7 @@
 -->
 <html lang="en">
 <head>
-    <title translatable="yes">Cockpit Starter Kit</title>
+    <title translate="yes">Cockpit Starter Kit</title>
```
mvollmer, line 20: Isn't just `<title translate>` enough? That would be best to show as an example, if correct.

mvollmer (changes requested):

src/index.html
martinpitt: I verified that it actually is, so I changed it accordingly.

martinpitt:
I also noticed that building the pot is totally broken:
```
❱❱❱ make po/starter-kit.pot
po/html2po -f po/POTFILES.html.in -o po/starter-kit.html.pot
po/manifest2po src/manifest.json -o po/starter-kit.manifest.pot
Option "--directory" requires 1 arguments, but 0 were provided
USAGE: node manifest2po [OPTION1] [OPTION2]... arg1 arg2...
The following options are supported:
  -d, --directory <ARG1> 	Base directory for input files
  -o, --output <ARG1>    	Output file
  -f, --from <ARG1>      	File containing list of input files
Option "--directory" requires 1 arguments, but 0 were provided
USAGE: node html2po [OPTION1] [OPTION2]... arg1 arg2...
The following options are supported:
  -d, --directory <ARG1> 	Base directory for input files
  -o, --output <ARG1>    	Output file
  -f, --from <ARG1>      	File containing list of input files
make: *** [Makefile:45: po/starter-kit.manifest.pot] Error 1
make: *** Waiting for unfinished jobs....
make: *** [Makefile:42: po/starter-kit.html.pot] Error 1
```

Fixed as well, and added to travis.

Makefile
```diff
@@ -39,10 +39,10 @@ po/POTFILES.html.in:
 	find src -name '*.html' > $@
 
 po/$(PACKAGE_NAME).html.pot: po/POTFILES.html.in
-	po/html2po -f $^ -o $@
+	po/html2po -d . -f $^ -o $@
 
-po/$(PACKAGE_NAME).manifest.pot:
-	po/manifest2po src/manifest.json -o $@
+po/$(PACKAGE_NAME).manifest.pot: src/manifest.json
+	po/manifest2po -d . -f $^ -o $@
```
martinpitt, line 45: Meh, this isn't right either. Why are these tools to utterly complicated??

martinpitt:
OK, now the generated pot file looks correct.

mvollmer (approved):

PR #382, opened 2020-10-17
Use sassc instead of node-sass

martinpitt:
node-sass is a compiled ELF module, which is problematic for
distributions that want to rebuild everything from source. The sassc CLI
program is packaged everywhere, and both use the same libsass library.

So drop node-sass and sass-loader, and replace it with a simple loader
wrapper around sassc.

 - [x] Land sassc in tasks/release containers: https://github.com/cockpit-project/cockpituous/pull/345

This also saves 122 npm packages (16 MB in node_modules/).


martinpitt:
This [requires sassc in tasks/release containers](https://github.com/cockpit-project/cockpituous/pull/345). I did that yesterday, but the PR did not actually land yet, so setting to blocked.

The generated CSS is not bitwise identical, as node-sass and Fedora 32's sassc use slightly different libsass versions. But the generated CSS differences seem marginal to me, and the page looks fine.

However, there is quite some difference in the generated css.map files, they are a lot smaller now. Therefore I'd like @garrett to sign off on this, as I don't know how to actually use them. Neither with master nor with this PR the Firefox CSS/style inspector looks any different, and I don't see any references to original CSS files either way.

martinpitt:
Oops, yes, [travis needs it as well](https://travis-ci.org/github/cockpit-project/starter-kit/builds/736593927).

martinpitt:
@lnussel: Is that the direction you had in mind? It was one of your points in the [recent cockpit packaging discussion](https://lists.fedorahosted.org/archives/list/cockpit-devel@lists.fedorahosted.org/thread/UJKOSUUKYBIY6QZ6IL5P455E63XYIAW5/). If that works and my colleagues agree, I'll apply the same to cockpit-{podman,ostree} and cockpit itself.

garrett:
I like this idea in general; sassc is basically the standard sass compiler that most people use. As far as maps go, I've never really gotten our maps properly working in Firefox anyway.

I'll look closer at this on Tuesday; I just wanted to chime in to say I approve of the idea and am looking forward to this.

> node-sass and Fedora 32's sassc use slightly different libsass versions

Which is using a newer sass library in F32? I'd guess differences are probably related to bug fixes or optimizations in the library. (I'll look closer soon.)

martinpitt:
@garrett: Nice to hear, thanks!

FTR, sassc in [Fedora 32 uses libsass 3.6.3](https://bodhi.fedoraproject.org/updates/FEDORA-2019-0deede1deb), whereas current node-sass uses and older 3.5.5:
``` node -p "require('node-sass').info"
node-sass	4.14.1	(Wrapper)	[JavaScript]
libsass  	3.5.5	(Sass Compiler)	[C/C++]
```

lnussel:
Sounds good to me. One less problem to worry about :)

garrett:
We might want to consider using dart-sass instead:
https://sass-lang.com/dart-sass & https://github.com/sass/dart-sass

It's the successor to libsass. It's where new features are being introduced and even supports nice things like `@import-once`.

"Dart Sass has replaced Ruby Sass as the canonical implementation of the Sass language."

Even though it's written in Dart, it's available in NPM too as "sass". However, it's slower at the moment in many cases (as it's transpiled Dart to native JS). So we might want to hold off and switch to sassc now and Dart Sass later... if it is slower for us. They're working on speeding it up, apparently.

martinpitt:
@garrett and I talked this over yesterday, and we thought that the best course for now is to move the "small" projects like starter-kit and c-podman to dart-sass right away -- these are easy enough to visually check for regressions. But let's do this approach for cockpit for now, and be a bit more conservative.

lnussel:
Has anyone tried to build Dart from source? It looks massive and I fear that would be the most complicated solution of all :/

martinpitt:
@lnussel : I haven't looked at dart-sass yet. My understanding was that the npm module was JavaScript, thus no ELF binaries. It's still transpiled though, so no preferred form of modification. So it would only solve the architecture portability side, not the license/rebuild from scratch problem. So maybe we should wait with dart-sass until that gets packaged into distros?

lnussel:
Sounds like C++ https://github.com/dart-lang/sdk/wiki/Building, there are also cross compiling instructions for ARM so looks pretty much like native code. So with that in mind I wonder if whatever is on npm can actually be the real thing. Maybe also just some precompiled js.

martinpitt:
Then I propose we review/land this as-is, use the approach for cockpit as well, and look at dart-sass separately (checking if it can be packaged into distros sensibly, and such). @garrett , @KKoukiou , want to have a look?

garrett (approved):
LGTM (as far as I can tell)

KKoukiou (approved):

PR #386, opened 2020-10-26
Run integration tests in Travis

martinpitt:
Travis now offers /dev/kvm, and its machines are powerful enough to run our browser integration tests, at least for small projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,31 @@
 dist: focal
-sudo: false
-language: node_js
-node_js:
-  - "lts/*"
+sudo: true
+language: minimal
 addons:
   apt:
     packages:
+      - appstream-util
+      - chromium-browser
+      - curl
+      - git
+      - libvirt-daemon-system
+      - npm
+      - python3-libvirt
+      - qemu-kvm
+      - qemu-utils
+      - rpm
       - sassc
+env:
+  - TEST_OS=fedora-33
+  - TEST_OS=centos-8-stream
 script:
-  - make
+  # HACK: /dev/kvm is root:kvm 0660 by default
+  - sudo chmod 666 /dev/kvm
+
+  # test PO template generation
   - make po/starter-kit.pot
+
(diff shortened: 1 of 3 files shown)


martinpitt:
I reworked this to actually build/use the rpm and thus replace the big hack with a lot smaller one.

I'd still like to move away from the huge `language: node_js` environment to a minimal one, to make things easier to reproduce locally and more explicit. But that could also be a follow-up.

martinpitt:
That was easier than I thought (success on first attempt), so here it is with `language: minimal`.

martinpitt:
Added `$TEST_JOBS`.

martinpitt:
Still fighting a bit with migrating this from legacy travis-ci.org to travis-ci.com, hang on..

martinpitt:
Ah, that may be the same trap that I had with github actions -- .travis.yml needs to be on master for GitHub to recognize/show the result. Anyway, the [run succeeded](https://travis-ci.com/github/martinpitt/starter-kit/builds/193087430).

marusak:
Needs rebase

martinpitt:
Rebased, and it works on Travis. But:

 - Integration tests on cockpit CI are now [broken](https://logs.cockpit-project.org/logs/pull-386-20201028-215557-dea68b1c-centos-8-stream/log.html) on a missing sassc. That sounds like PR #382 somehow broke this in conjunction with the recent container rebuild? The most recent tasks container was built after tests in #382 ran.

 - Travis results don't appear here. .travis.yml is already on master, but somehow the integration is broken. This needs investigating.

Moving to draft for now.

martinpitt:
I rebuilt the tasks container from current cockpituous master, which brought back sassc. Cockpit CI tests succeed now.

martinpitt:
Ah, c-p/starter-kit was still registered in the old travis-ci.org, and I disabled it there. Migrated over to travis-ci.com, and it properly appears here now. :tada:

marusak (approved):
Nice, thanks!

PR #395, opened 2020-11-22
Webpack cleanups

martinpitt:


martinpitt:
FTR, I tried to move to [webpack 5](https://webpack.js.org/migrate/5/), but it's not worth it -- the built webpacks are bigger, and the build takes longer. Still nice to have some cleanups, though.

martinpitt:
@KKoukiou : Do you have some time to look at this? Should be fairly easy, I hope. Epharisto!

KKoukiou (approved):
Quite a massive simplification, thanks!

PR #396, opened 2020-11-25
Bump test/common dependency to 233

martinpitt:
This gets rid of the hack for
<https://github.com/cockpit-project/cockpit/issues/14133>.


marusak (approved):
Thanks!

PR #406, opened 2020-12-17
spec: Add BuildRequires: make

martinpitt:
See https://fedoraproject.org/wiki/Changes/Remove_make_from_BuildRoot


marusak (approved):
Nice, thanks!
Also when you install Fedora it does not include `make` either, so this would help newcomers as well.

PR #408, opened 2020-12-26
Translation cleanups

martinpitt:
I eventually want to simplify this further and build the files straight through webpack. This was some cleanup fallout from me getting more familiar with how our translation system works.


marusak (approved):
Thanks :)

PR #409, opened 2020-12-28
workflows: Move npm-update to Ubuntu 20.04

martinpitt:
GitHub's 18.04 additional repositories break NPM. As "ubuntu-latest" is
going to switch to 20.04 soon anyway [1], do the jump now.

[1] https://github.com/actions/virtual-environments/issues/1816

----

This fixes the recent [npm-update failures](https://github.com/cockpit-project/starter-kit/actions). I [tested this on my fork](https://github.com/martinpitt/starter-kit/runs/1616426454?check_suite_focus=true). Once this lands, we need to apply the same change to -podman, -ostree, and friends.

### DIFF
--- a/.github/workflows/npm-update.yml
+++ b/.github/workflows/npm-update.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 jobs:
   npm-update:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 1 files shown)


martinpitt:
Note: We could alternatively also let this run in a cockpit/tasks container, but this is a less intrusive change for now.

Gundersanne (approved):

PR #410, opened 2020-12-30
Build translations with webpack

martinpitt:
Convert the `po2json` script into a webpack plugin, and integrate the
(now trivial) po.empty.js template.

This is the last step for building the entire dist/ directory with
`npm run build` (i.e. a tool that web developers are familiar with),
and not having a split webpack+make toolchain any more.

 - [x] Builds on top of PR #408


martinpitt:
I originally tried to [implement this with a webpack loader](https://github.com/cockpit-project/starter-kit/files/5753828/starter-kit-po-webpack-loader.patch.txt), which seemed more elegant at first. But webpack adds its bootstrapping/module code to all the produced .js files. This unnecessarily blows up the files, but more importantly they break the `<script type="text/javascript" src="../*/po.js">` inclusion as webpack's bundles lack a final newline. This breaks the syntax of concatenated files. That little problem is so annoyingly hard to fix, that I went for a plugin instead, which lets us have full control over the created files.

I diffed the files before and after this PR, and they are identical except for the comment.

marusak:
#408 loaded

martinpitt:
Rebased.

KKoukiou (approved):

PR #411, opened 2020-12-30
Adjust test for changed display language modal

martinpitt:
https://github.com/cockpit-project/cockpit/commit/f1542d54dba rewrote
the language selector dialog. Update the tests to also work with
cockpit ≥ 233.

Do the version comparison dynamically for the time being, as this breaks
several image refreshes. We can drop this again and only use the new
code once all of these landed.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -36,8 +36,14 @@ class TestApplication(testlib.MachineCase):
         b.click("#navbar-dropdown")
         b.click(".display-language-menu a")
         b.wait_popup('display-language')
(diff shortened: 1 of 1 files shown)


martinpitt:
Oops, [fractional version numbers](https://logs.cockpit-project.org/logs/pull-411-20201230-161458-f8df94ee-centos-8-stream/log.html). Fixed.

marusak (approved):
> Oops, fractional version numbers. Fixed.

Heh, I was about to comment why not using `int` :) Makes sense, thanks!

PR #413, opened 2021-01-07
cockpit-po-plugin: Fix crash with plural forms

martinpitt:
Declare the `pos` variable. Fixes regression introduced in commit
7f6ef51c12a7.

Add plural forms to German translations to exercise this code path.


marusak (approved):
Good catch!
Would tests now fail when just introducing change into `po/de.po`?

martinpitt:
@marusak: Yes, the build fails, webpack crashes on `pos is not defined`. I noticed this when backporting this stuff to c-podman in https://github.com/cockpit-project/cockpit-podman/pull/645 (that PR has this fix included already).

PR #414, opened 2021-01-18
Use Cockpit webpack helpers from cockpit.git

martinpitt:
Drop the cockpit-po-plugin.js and sassc-loader.js code copies, and check
them out from a stable cockpit tag, similarly to how we already get the
PatternFly CSS.

 - [x] Fix `cockpit-po-plugin.js` without options: https://github.com/cockpit-project/cockpit/pull/15166
 - [x] Wait for next cockpit release, checkout stable 236 tag instead of master


martinpitt:
cockpit 236 exists now, so I switched this to that stable tag. Ready to land now.

KKoukiou (approved):

PR #416, opened 2021-01-25
Run tests on Cirrus CI, drop Travis

martinpitt:
Travis stopped handing out free credits since their plan changes last
November, so tests have not run since then.

Cirrus CI offers an API similar to Travis, and also explicitly supports
/dev/kvm access. They run on the GKE Linux Community Cluster, and thus
are not as likely to run into monetary problems as Travis did.

Eventually we want to use the cockpit/tasks container directly. But we
can't fix the /dev/kvm permissions inside that as `sudo` does not work
in that. Once we fix that, the setup will become much cheaper and
simpler.


martinpitt:
I developed this on my fork, and my [most recent run succeeded](https://cirrus-ci.com/build/6066177044643840). I now enabled Cirrus CI for cockpit-project, but only for starter-kit for now. We can extend it to other projects as we see fit, and if there are too many, just enable it for all repos for simplicity (it will only cover projects with an actual `.cirrus.yml`).

martinpitt:
The [graph](https://cirrus-ci.com/task/5459960264916992) clearly shows that we don't need 2 CPUs, let's set it to just one.

marusak (approved):
Very nice and elegant! thanks
(one core seems to be enough)

martinpitt:
@marusak : Right, the current version just uses one core.

TomasTomecek (commented):
so impressive that you switched from one provider to another "just like that"

martinpitt:
@TomasTomecek : <SteveBallmer>Containerize! Containerize! Containerize! :grinning:

PR #419, opened 2021-02-01
Enable webpack watch progress

martinpitt:
Without that, `npm run watch` is entirely silent after the first build,
and does not give any feedback when a build starts, and even more
importantly, when it's done.


martinpitt:
@marusak : FYI, this is what I talked about this morning. Once this lands, I'd like to port it to podman as well.

KKoukiou (approved):

PR #421, opened 2021-02-08
Introduce a github action which builds a cockpit preview when requested

KKoukiou:
It currently reacts to /livetest comments on PRs and builds cockpit from
the PR's checkout. Then serves it on logs.cockpit-project.org on a port
specified on a follow-up comment to the /livetest.

Cockpit developers can now utilize this workflow to review a PR without
having to pull and compile the code locally.

You can try it out here: https://github.com/KKoukiou/starter-kit/pull/1

 - [x] Requires SSH gateway setup in our infra: https://github.com/cockpit-project/cockpituous/pull/381

### DIFF
--- a/.github/workflows/livetest.yml.disabled
+++ b/.github/workflows/livetest.yml.disabled
@@ -1,0 +1,99 @@
+name: Live test cockpit from PRs
+# Triggered when a repository collaborator comments /livetest on a PR
+# This is currently configured to run on cockpit-project org.
+# If you need to set up your own relay host and adjust the RELAY_HOST env variable
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  pr-info:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Query author repository permissions
+        uses: octokit/request-action@v2.x
+        id: user_permission
+        with:
+          route: GET /repos/${{ github.repository }}/collaborators/${{ github.event.sender.login }}/permission
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # restrict /livetest to users with admin or write permission for the repository
+      # see https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-repository-permissions-for-a-user
+      # store output if user is allowed in allowed_user job output so it has to be checked in downstream job
+      - name: Check if user does have correct permissions
+        if: contains('admin write', fromJson(steps.user_permission.outputs.data).permission)
+        id: check_user_perm
+        run: |
+          echo "User '${{ github.event.sender.login }}' has permission '${{ fromJson(steps.user_permission.outputs.data).permission }}' allowed values: 'admin', 'write'"
+          echo "::set-output name=allowed_user::true"
+
+    outputs:
+      allowed_user: ${{ steps.check_user_perm.outputs.allowed_user }}
+
+  live-test:
+    needs: pr-info
+    if: needs.pr-info.outputs.allowed_user == 'true' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/livetest')
+    runs-on: ubuntu-20.04
+    env:
+      RELAY_HOST: logs.cockpit-project.org
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y cockpit-system cockpit-ws nodejs npm sassc make
+
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          # need this to also fetch tags
+          fetch-depth: 0
+
+      - name: Compile and install the project
+        run: make && sudo make install
+
+      - name: Create a new test user
+        # WARNING: if you want to add this users to sudoers make sure to propertly delete the SINK_SSHRELAY_PRIVATE
+        run: sudo useradd --create-home livetest; echo livetest:foobar | sudo chpasswd
+
+      - name: Get the commit SHA that triggered the workflow
+        id: get_commit_sha
+        run: |
+          echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
+      - name: Get ssh gateway port to forward to
+        id: remote_port
+        env:
+          SHA_SHORT: "${{ steps.get_commit_sha.outputs.sha_short }}"
+        run: |
+            echo ::set-output name=SSH_GATEWAY_PORT::$((16#$SHA_SHORT % 32768))
+
+      - name: Write a comment on the PR with link to the cockpit instance
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Cockpit preview: https://${{ env.RELAY_HOST }}:${{ steps.remote_port.outputs.SSH_GATEWAY_PORT }}'
+            })
+
+      - name: Create an SSH private key from the secret
+        run: |
+          sudo touch /run/SINK_SSHRELAY_PRIVATE && sudo chown $(id -u):$(id -g) /run/SINK_SSHRELAY_PRIVATE
+          echo "${{ secrets.SINK_SSHRELAY_PRIVATE }}" > /run/SINK_SSHRELAY_PRIVATE
+          chmod 600 /run/SINK_SSHRELAY_PRIVATE
+          mkdir -p ~/.ssh &&  ssh-keyscan -H ${{ env.RELAY_HOST }} > ~/.ssh/known_hosts
+
+      - name: Forward cockpit port 9090 to a server we can access
+        run: ssh -f -R ':${{ steps.remote_port.outputs.SSH_GATEWAY_PORT }}':localhost:9090 -i /run/SINK_SSHRELAY_PRIVATE -o ExitOnForwardFailure=yes sshgw@${{ env.RELAY_HOST }} 'sleep infinity'
+
+      - name: Block workflow from finishing
+        run: sleep 4h
+
+      - name: Cleanup ssh process
+        if: ${{ always() }}
+        run: pkill ssh || true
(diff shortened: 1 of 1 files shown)


.github/workflows/livetest.yml
```diff
@@ -0,0 +1,62 @@
...
+          mkdir -p ~/.ssh &&  ssh-keyscan -H logs.cockpit-project.org > ~/.ssh/known_hosts
+
+      - name: Forward cockpit port 9090 to a server we can access
+        run: ssh -tt -R ':${{ steps.remote_port.outputs.AWS_PORT_NUMBER }}':localhost:9090  -i /tmp/SINK_SSHRELAY_PRIVATE sshgw@logs.cockpit-project.org 'sleep infinity' &
+
+      - name: Delete security keys
+        run: rm /tmp/SINK_SSHRELAY_PRIVATE
+
+      - name: Block workflow from finishing
+        run: sleep infinity
```
martinpitt, line 62: Hmm, I wonder if we should have a better way of ending it, as like this it will always time out and generate an error email. The simplest would probably to `sleep 4h`.

martinpitt (changes requested):
Amazing,  thank you! My usual round of nitpicking, but I'm really looking forward to putting this into action with cockpit itself.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,62 @@
...
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Cockpit preview: https://logs.cockpit-project.org:${{ steps.remote_port.outputs.AWS_PORT_NUMBER }}'
+            })
+
+      - name: Create an SSH private key from the secret
+        run: |
+          echo "${{ secrets.SINK_SSHRELAY_PRIVATE }}" > /tmp/SINK_SSHRELAY_PRIVATE
```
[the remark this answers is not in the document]
KKoukiou: I left it on /run just chown one file, hope it's fine.

martinpitt:
I'm a bit surprised that you have a KKoukiou:livetest branch for this. I thought you were developing this on your fork's master branch? Did you keep the two in sync? (Note that it's perfectly possible to propose a PR from fork/project:master to cockpit-project/project/:master) I.e. can we try the current version on your test PR on your fork?

KKoukiou:
> I'm a bit surprised that you have a KKoukiou:livetest branch for this. I thought you were developing this on your fork's master branch? Did you keep the two in sync? (Note that it's perfectly possible to propose a PR from fork/project:master to cockpit-project/project/:master) I.e. can we try the current version on your test PR on your fork?

I just synced my forks master with this branch and created a PR. https://github.com/KKoukiou/starter-kit/compare/master...KKoukiou:livetest 

Just out of the habbit to not created PRs from master, not any reason behind it.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,87 @@
...
+          sudo touch /run/SINK_SSHRELAY_PRIVATE && sudo chown $(id -u):$(id -g) /run/SINK_SSHRELAY_PRIVATE
+          echo "${{ secrets.SINK_SSHRELAY_PRIVATE }}" > /run/SINK_SSHRELAY_PRIVATE
+          chmod 600 /run/SINK_SSHRELAY_PRIVATE
+          mkdir -p ~/.ssh &&  ssh-keyscan -H logs.cockpit-project.org > ~/.ssh/known_hosts
+
+      - name: Forward cockpit port 9090 to a server we can access
+        run: ssh -R ':${{ steps.remote_port.outputs.SSH_GATEWAY_PORT }}':localhost:9090  -i /run/SINK_SSHRELAY_PRIVATE sshgw@logs.cockpit-project.org &
+
+      - name: Delete security keys
+        run: sudo rm /run/SINK_SSHRELAY_PRIVATE
```
martinpitt, line 84: This is a race condition -- the file may get removed before ssh gets to reading it. Perhaps we could fix this with using `ssh -f`?

martinpitt: This is also interesting:

> If the ExitOnForwardFailure configuration option is set to “yes”, then a client started with -f will wait for all remote port forwards to be successfully established before placing itself in the background.

You can set that with `-o ExitOnForwardFailure=yes`

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          # need this to also fetch tags
+          fetch-depth: 0
+
+      - name: Compile and install the project
+        run: make && sudo make install
+
+      - name: Create a new test user
+        # WARNING: if you want to add this users to sudoers make sure to propertly delete the SINK_SSHRELAY_PRIVATE
+        run: sudo useradd livetest; echo livetest:foobar | sudo chpasswd
```
martinpitt, line 53: Can you please add `--create-home`? Debian/Ubuntu's `useradd` does not do that by default, and if you try to use the Terminal in cockpit it just fails on "no such file or directory".

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,89 @@
...
+          sudo touch /run/SINK_SSHRELAY_PRIVATE && sudo chown $(id -u):$(id -g) /run/SINK_SSHRELAY_PRIVATE
+          echo "${{ secrets.SINK_SSHRELAY_PRIVATE }}" > /run/SINK_SSHRELAY_PRIVATE
+          chmod 600 /run/SINK_SSHRELAY_PRIVATE
+          mkdir -p ~/.ssh &&  ssh-keyscan -H logs.cockpit-project.org > ~/.ssh/known_hosts
+
+      - name: Forward cockpit port 9090 to a server we can access
+        run: ssh -f -R ':${{ steps.remote_port.outputs.SSH_GATEWAY_PORT }}':localhost:9090 -i /run/SINK_SSHRELAY_PRIVATE -o ExitOnForwardFailure=yes sshgw@logs.cockpit-project.org 'sleep infinity'
+
+      - name: Block workflow from finishing
+        run: sleep 4h
```
martinpitt, line 85: FTR: This actually works well even with cancelling -- I just did that, and the workflow *did* run the "cleanup ssh" step, and I just checked on the sink that there are no leftover processes. :+1:

martinpitt (changes requested):
Amazing, thank you! I just tested this end-to-end, and it's a great experience.

The only thing I'm not sure about is whether we actually want to land that in starter-kit as-is. Maybe as `.yml.disabled` like we have the "release" workflow? With a comment that you need to set up your own relay server? If you prefer to land it, then this should be documented in README.md.

But for sure the next step would be to apply this to cockpit-podman?

.github/workflows/livetest.yml
KKoukiou: added!

KKoukiou:
> Amazing, thank you! I just tested this end-to-end, and it's a great experience.
> 
> The only thing I'm not sure about is whether we actually want to land that in starter-kit as-is. Maybe as `.yml.disabled` like we have the "release" workflow? With a comment that you need to set up your own relay server? If you prefer to land it, then this should be documented in README.md.
> 
> But for sure the next step would be to apply this to cockpit-podman?

I just disabled it for now.

.github/workflows/livetest.yml
KKoukiou: Super! Problably as a followup we would like to add a way for the user to cancel that, in case the PR was repushed and they want to regenerate the preview.

martinpitt: That shouldn't collide, though? You can easily have many workflows running in parallel. That's a waste, but not broken.

KKoukiou: Sure, but only one ssh tunnel will be active in the port that the PR corresponds to with the current implementation.

.github/workflows/livetest.yml.disabled
```diff
@@ -0,0 +1,89 @@
+name: Live test
```
martinpitt, line 1: Let's add a comment what this does and what is needed to enable it (run in cockpit-project org, or you need to set up your own relay host and adjust the `ssh` command in the "Forward cockpit port 9090" step)

martinpitt (commented):
Cheers!

.github/workflows/livetest.yml
martinpitt: Oh, of course, the port would collide. This can be fixed later by not taking the issue number, but the commit SHA % 32768 (so that it fits into the available port range and uses a wide range).

KKoukiou: I will make this adjustment now - it makes sense that we don't expect the user to stop the workflow to start a new one.

.github/workflows/livetest.yml.disabled
KKoukiou: I will make the RELAY host an env variable and add the comment.

KKoukiou:
@martinpitt just tested in on master and work as expected. The only difference between master and this branch is that it's not disabled there.

.github/workflows/livetest.yml.disabled
```diff
@@ -0,0 +1,99 @@
...
+        id: get_commit_sha
+        run: |
+          echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
+      - name: Get ssh gateway port to forward to
+        id: remote_port
+        env:
+          SHA_SHORT: "${{ steps.get_commit_sha.outputs.sha_short }}"
+        run: |
+            echo ::set-output name=SSH_GATEWAY_PORT::$((16#$SHA_SHORT % 32768))
```
martinpitt, line 70: Please still add "+ 1500", otherwise this has a chance of picking a low port; user processes are not allowed to open ports < 1024, and it may also collide with ssh or http.

.github/workflows/livetest.yml.disabled
```diff
@@ -0,0 +1,99 @@
+name: Live test cockpit from PRs
+# Triggered when a repository collaborator comments /livetest on a PR
+# This is currently configured to run on cockpit-project org.
+# If you need to set up your own relay host and adjust the RELAY_HOST env variable
```
martinpitt, line 4:
```suggestion
# If your project lives somewhere else than the github cockpit-project organization,
# you need to set up your own relay host, add a secret with the private SSH key, and
# adjust the RELAY_HOST env variable and SINK_SSHRELAY_PRIVATE secret name.
```

martinpitt (changes requested):
Thanks! I see a VM running in https://github.com/KKoukiou/starter-kit/pull/1 but I can't reach it, hmm..

martinpitt (approved):
Cheers! Please squash the commits on landing.

PR #425, opened 2021-02-19
test: Generate screenshot artifact of main screen

martinpitt:
This is a robust screenshot of a steady state without expected jitter.

This is useful for visually validating PatternFly update PRs without
having to locally build and run. In the future we can also run
programmatic image diffs on these to auto-detect pixel
changes/layout regressions.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -25,6 +25,7 @@ class TestApplication(testlib.MachineCase):
         # verify expected host name
         hostname = m.execute("hostname").strip()
         b.wait_in_text(".pf-c-alert__title", "Running on " + hostname)
+        b.snapshot("main-english")
 
         # change current hostname
         m.execute("echo new-%s > /etc/hostname" % hostname)


martinpitt:
Nice, so now the successful tests have that screenshot artifact which is easily available.

@mvollmer , WDYT? Is that roughly the approach that we want to take in cockpit? Or did you have something else in mind?

If we want this, I'd like to refine it a bit:

  * Only do the screenshot artifacts on `$TEST_OS_DEFAULT`, to avoid too much artifact upload/storage overhead
  * Don't capture a corresponding .html artifact, that's quite pointless for this purpose.
  * The above warrant a new convenience function in testlib.

mvollmer:
> WDYT? Is that roughly the approach that we want to take in cockpit?

Yes, that's what I had in mind.

martinpitt:
Closing in favor of @mvollmer 's pixel testing PRs.

PR #426, opened 2021-02-24
Fix repo clone in livetest workflow

KKoukiou:


martinpitt (approved):
Thanks! Feels a bit complicated, it feels like there should be an easier way to do that. But it's fine, we can tweak that later.

PR #427, opened 2021-02-25
Drop unused version numbers from package.json and manifest

### LINKED ISSUE #200: version number needs to be stored in a single location
There are different version numbers stored in:

- package.json, "0.1.0"
- manifest.json "0.1" 
- cockpit-starter-kit.spec.in "1%{?dist}"

Which one is correct?

There should be a single source of truth i.e. the version number should be stored in only one location

martinpitt:
This field was introduced in
https://github.com/cockpit-project/cockpit/pull/4964 as "purely
informational for now", and isn't even parsed by cockpit.

package.json's version would only be relevant for publishing NPM
modules, but cockpit pages are not that.

Neither starter-kit itself nor our derived projects like cockpit-podman
or cockpit-composer have ever maintained these two fields, so just get
rid of them.

This makes the git tag the single source of truth for the version
number.

Fixes #200

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "name": "starter-kit",
-  "version": "0.1.0",
   "description": "Scaffolding for a cockpit module",
   "main": "index.js",
   "repository": "git@github.com:cockpit/starter-kit.git",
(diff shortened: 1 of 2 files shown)


martinpitt:
FTR, I'll apply this to cockpit-{podman,composer,ostree} once this landed.

KKoukiou (approved):

PR #428, opened 2021-02-25
Bump cockpit component library to 238

martinpitt:
Don't land that yet, I want to use it to test f-34 as well.


martinpitt:
Argh, our [run-tests regressed](https://logs.cockpit-project.org/logs/pull-428-20210225-070902-367fa387-fedora-33/log.html) for non-cockpit. I'll look at that separately, and only bump the shared component version for now.

marusak (approved):

PR #432, opened 2021-03-04
Add FMF tests, run in packit

martinpitt:


martinpitt:
Nice, this works! packit seems to be a bit slow/stuck, I asked in https://github.com/cockpit-project/cockpit-podman/pull/687 (which has the same problem). But rawhide worked here.

marusak (approved):
Thanks!

PR #433, opened 2021-03-05
Pixel testing experiment

mvollmer:


mvollmer:
:-( No PIL in the test container...

mvollmer:
> :-( No PIL in the test container...

Fixed.

mvollmer:
Oho, this is expected to pass, but "gastgebername" fails with these differences:

image

(I had to write special code to mark them red, the differences are impossible to spot with image github image comparison)

mvollmer:
> This is a bit discouraging already...

It seems to flip between two versions... here is a interesting run with two passes and one fail: https://logs.cockpit-project.org/logs/pull-433-20210312-125319-b84786d7-fedora-33/log.html

mvollmer:
The plan: allow reasonable differences in intensity, but not in color.

PR #435, opened 2021-03-12
Pixel Testing Experiment Test PR

mvollmer:
This PR changes the looks of the UI and the pixel tests that have been added in #436 catch that, as expected.  The next step is to update the pixel test references to confirm that this is intended. See #438 for this.


mvollmer:
This is referenced from https://mvollmer.github.io/pixel-test-log-demo/ and still interesting. But let's formally close it since we will never merge it and we don't want to clutter up our lists.

PR #436, opened 2021-03-18
Pixel testing experiment

mvollmer:
This PR adds two pixel test points and their reference images (in a submodule).  See #435 for a PR that makes a change that is caught by these new tests.


mvollmer:
Test

mvollmer:
This is referenced from https://mvollmer.github.io/pixel-test-log-demo/ and still interesting. But let's formally close it since we will never merge it and we don't want to clutter up our lists.

PR #438, opened 2021-03-18
Pixel testing experiment Test PR v2

mvollmer:
This PR makes the same visual changes as #435, but also updates the pixel test references.  Thus, the test is green and we can review the changes to the look here in this PR.


mvollmer:
This is referenced from https://mvollmer.github.io/pixel-test-log-demo/ and still interesting.  But let's formally close it since we will never merge it and we don't want to clutter up our lists.

PR #439, opened 2021-03-31
Pixel Testing Experiment (without reference pixels)

mvollmer:
This is like #436, but the reference pixels are not yet included.


mvollmer:
Whoops, wrong base.

PR #447, opened 2021-04-09
Build and ship LICENSE file

martinpitt:
Bring back terser-webpack-plugin from commit c52136c051448 and re-enable
it. It builds the autogenerated dist/index.js.LICENSE.txt which is quite
important for shipping distribution packages -- c-{podman,machines} both
do this.

Backport https://github.com/cockpit-project/cockpit-podman/commit/95021c336ed
to make sure that the webpack for the release tarball is *actually*
built in production mode.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,6 +3,7 @@ const childProcess = require('child_process');
 
 const copy = require("copy-webpack-plugin");
 const extract = require("mini-css-extract-plugin");
+const TerserJSPlugin = require('terser-webpack-plugin');
 const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CompressionPlugin = require("compression-webpack-plugin");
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -69,7 +70,7 @@ module.exports = {
 
     optimization: {
         minimize: production,
-        minimizer: [`...`, new CssMinimizerPlugin()],
+        minimizer: [new TerserJSPlugin(), new CssMinimizerPlugin()],
```
KKoukiou, line 73: This actually should be the default according to https://webpack.js.org/configuration/optimization/#optimizationminimizer 
At least that what I understood when I reviewed the initial PR.

KKoukiou: So I just tried it now locally, and indeed it does not work without so I am happy with the fix. Too bad that the webpack docs are a but fuzzy (to me at least).

KKoukiou (approved):

PR #473, opened 2021-07-06
test: Fix host name lookup

martinpitt:
Our code reads /etc/hostname. Do the same in the test, as the runtime
sethostname(2) (set by hostnamed) and static host name may be different.


allisonkarlitskaya (approved):
Uh.  Sure.

Once day when we're old and grey, maybe you'll explain to me what a hostname is.  I have to admit, I don't understand it, myself, anymore.

PR #488, opened 2021-09-09
Rebuild webpack during RPM build, move to `make dist` xz tarballs

martinpitt:
This is necessary to comply with Fedora's packaging policy:
https://docs.fedoraproject.org/en-US/packaging-guidelines/JavaScript/
    
Include the node cache in the source rpm, unpack it into the main source
dir, and force a webpack rebuild in `%build`.

-----

Ported from https://github.com/skobyda/cockpit-certificates/pull/65 . It's generally a nice life insurance to release the `node_modules/` together with the software -- then distros can still patch old versions easily.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *~
 *.retry
-*.tar.gz
+*.tar.xz
 *.rpm
 node_modules/
 dist/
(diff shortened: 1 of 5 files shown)


martinpitt:
After pondering this for a bit, I applied the full thing. IMHO it's more correct to provide the Fedora standards-compatible way  in starter-kit by default. It's easier to opt out of this than discovering how to opt in.

KKoukiou (approved):

PR #490, opened 2021-09-13
.gitignore: add tmp directory

jelly:
Include to tmp/ directory to be ignored, it usually contains temporary
vm-run images.


martinpitt (approved):
Danke!

PR #496, opened 2021-09-21
webpack: Avoid md4 hash for OpenSSL 3 compatibility

martinpitt:
CentOS/RHEL 9 switched to OpenSSL 3, which does not support the `md4`
hash any more. webpack 5 hardcodes that in no less than 21 places, so
monkey-patch `crypto.createHash()` to substitute sha256 for md4.

Hack to work around https://github.com/webpack/webpack/issues/13572


martinpitt:
This should fix [these rawhide build failures](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-starter-kit-494/fedora-rawhide-x86_64/02828853-cockpit-starter-kit/builder-live.log.gz), which we started to get ≤ 4 days ago.

marusak (approved):

PR #500, opened 2021-10-07
FMF/packit test improvements

martinpitt:
Avoids the chromium crashes in [Fedora 34](http://artifacts.dev.testing-farm.io/663e2adf-1858-487d-9232-6c3255c3d5e8/) and [rawhide](http://artifacts.dev.testing-farm.io/96e0f761-dda8-4017-9445-1d6d15b44df5/), and syncs some improvements from our other projects.

### DIFF
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -5,7 +5,7 @@ require:
   - cockpit-ws
   - cockpit-system
   - bzip2
-  - git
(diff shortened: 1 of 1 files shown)


marusak (dismissed):
Seems to work, thanks!
Just one comment, but not necessary to re-run all tests just for that.

PR #513, opened 2021-11-29
test: Drop obsolete language switcher cases

martinpitt:
The oldest release which we support is 251 from RHEL/CentOS 8.5, so we
don't need the older cases any more. Also move to
MachineCase.system_before().


marusak (approved):
Thanks!

PR #519, opened 2021-12-15
fmf: Go back to firefox distro version

martinpitt:
By now the packaged firefox contains CDP. This makes the tests more
robust against CDP breakage in nightly.

----

Currently *all* packit and downstream gating tests are [broken](http://artifacts.dev.testing-farm.io/33d67a87-50b8-43bf-815b-aa81b696ab10/) , as last night, firefox nightly broke `frameNavigated()`/`executionContextCreated`. Fedora 35's browser is fine, let's check on the others.

If that doesn't work, https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=linux64&lang=en-US also works ok.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -12,12 +12,9 @@ LOGS="$(pwd)/logs"
 mkdir -p "$LOGS"
 chmod a+w "$LOGS"
 
-# install firefox (available everywhere in Fedora and RHEL); install the package to pull in all the dependencies
+# install firefox (available everywhere in Fedora and RHEL)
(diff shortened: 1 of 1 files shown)


martinpitt:
The rawhide packit test failure is different, it's already been around since two days ago (also still on my list, if other stuff would not keep popping up..). So let's land this ASAP, and I'll apply this everywhere else.

mvollmer (approved):

martinpitt:
I reported this upstream: https://bugzilla.mozilla.org/show_bug.cgi?id=1746154

PR #521, opened 2022-01-04
Add ability to rsync artifacts to a remote machine

jelly:
Use the CockpitRsyncWebpack plugin to optionally rsync build code to a
remote machine, the destination is determined by looking at the
package.json's name field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ or
 
     $ make watch
 
+When developing against a virtual machine, webpack can also automatically upload
+the code changes by setting the `RSYNC` environment variable to
+the remote hostname.
+
+    $ RSYNC=c make watch
+
 # Running eslint
 
 Cockpit Starter Kit uses [ESLint](https://eslint.org/) to automatically check
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,3 +1,4 @@
+const fs = require("fs");
 const path = require("path");
 
 const copy = require("copy-webpack-plugin");
@@ -7,9 +8,13 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CompressionPlugin = require("compression-webpack-plugin");
 const ESLintPlugin = require('eslint-webpack-plugin');
 const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
+const CockpitRsyncPlugin = require("./src/lib/cockpit-rsync-plugin");
 
 /* A standard nodejs and webpack pattern */
 const production = process.env.NODE_ENV === 'production';
+
+// Obtain package name from package.json
+const packageJson = JSON.parse(fs.readFileSync('package.json'));
 
 // Non-JS files which are copied verbatim to dist/
 const copy_files = [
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -43,6 +43,10 @@ or
 
     $ make watch
 
+When developing against a virtual machine webpack can also automatically upload
+the code changes automatically by setting the `RSYNC` environment variable to
+the remote hostname.
```
martinpitt, line 48: Drop the second "automatically"

webpack.config.js
```diff
@@ -7,10 +8,15 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CompressionPlugin = require("compression-webpack-plugin");
 const ESLintPlugin = require('eslint-webpack-plugin');
 const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
+const CockpitRsyncPlugin = require("./src/lib/cockpit-rsync-plugin");
 
 /* A standard nodejs and webpack pattern */
 const production = process.env.NODE_ENV === 'production';
 
+// Obtain package name from package.json
+const data = fs.readFileSync('./package.json', { encoding:'utf8', flag:'r' });
```
martinpitt, line 17: Let's add spaces after `:`.

But I don't think you'll need these options at all -- e.g. [here](https://github.com/cockpit-project/cockpit/blob/main/.github/workflows/npm-install.yml#L90) we read package.json without any options.

Please either eliminate the `data` const name (preferred), or rename it to `package_json_str` or so.

README.md
```diff
@@ -43,6 +43,10 @@ or
...
+
```
martinpitt, line 49: Perhaps give an example here?

    make watch RSYNC=c

martinpitt (changes requested):
Nice!

README.md
```diff
@@ -43,6 +43,10 @@ or
 
     $ make watch
 
+When developing against a virtual machine webpack can also automatically upload
```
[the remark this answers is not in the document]
martinpitt: This missing comma is still relevant in your latest force-push. In this case it makes the sentence quite a bit more readable.

webpack.config.js
```diff
@@ -7,10 +7,14 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CompressionPlugin = require("compression-webpack-plugin");
 const ESLintPlugin = require('eslint-webpack-plugin');
 const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
+const CockpitRsyncPlugin = require("./src/lib/cockpit-rsync-plugin");
 
 /* A standard nodejs and webpack pattern */
 const production = process.env.NODE_ENV === 'production';
 
+// Obtain package name from package.json
+const packageJson = JSON.parse(require('fs').readFileSync('package.json'));
```
martinpitt, line 16: For consistency, please move the require() back to the top. This was fine for an one-liner in a github workflow, but not for a full program. We may need to use it in another place in the future, and don't want to import it more than once.

martinpitt (approved):
:heart_eyes_cat: thanks!

PR #524, opened 2022-02-04
Various small packaging/npm updates

martinpitt:
Backported from recent cockpit-podman etc., nothing surprising.

I also have a commit for using `image-customize --build`, but that does not work yet -- starter-kit rebuilds the webpack during rpm build, so we need to add nodejs to our mocks on our images first.


martinpitt:
New copy-webpack-plugin causes [trouble on TF c8s](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-starter-kit-524/centos-stream-8-x86_64/03294580-cockpit-starter-kit/builder-live.log.gz). Curiously it worked in Cirrus CI and our own images..

marusak (approved):
Thanks!

PR #525, opened 2022-02-07
Use image-customize --build to build/install distribution packages

martinpitt:
This centralizes/factorizes the rpm package builds, and builds
RPM packages in the VM instead of on the host, which is cleaner. It also
paves the way for supporting Debian and arch builds.

Drop the `make srpm` rule, as it's not very useful. Keep the `make rpm`
rule, as sometimes developers do this manually. This might be replaced
later on with another image-customize feature which copies the built rpm
out of the VM.

----

Depends on https://github.com/cockpit-project/bots/pull/2903

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ dist/
 /.vagrant
 package-lock.json
 Test*FAIL*
-bots/
+/bots
 test/common/
 test/images/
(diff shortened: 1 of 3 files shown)


martinpitt:
@thrix : Still the same problem :cry:

PR #527, opened 2022-02-10
github: replace npm-update workflow with builtin dependabot [1] npm integration

KKoukiou:
[1] https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically


KKoukiou:
Anaconda does it like this and I think it's nicer than what we were doing. What do you think?

martinpitt:
How does such a generated PR look like? Our npm-update script has some logic to group related packages together (like `@patternfly/*`, but otherwise send separate PRs for every module. That makes debugging a lot easier. If it sends separate PRs for each module, that's going to break PF; if it updates everything at once, that feels too much.

KKoukiou:
Here is my PRs once I enabled it https://github.com/KKoukiou/starter-kit/pulls

KKoukiou:
@martinpitt what you describe (about groupped updates for patternfly packages) seems to be an open issue https://github.com/dependabot/dependabot-core/issues/1296.

I wonder if its' a blocker however. Will investigate a bit more.

KKoukiou:
I put some extra thought on this: 
* There is not dependency between @patternfly/react-* packages and @patternfly/patternfly. The latter is just a devDependency of the @patternfly/react-styles. For this reason I don't expect dependabot to group @patternfly/patternfly with anything, and also there is no need for it.

`npm ls` on cockpit-* repos, should produce something like this, excluding the versions. 

```
...
├── @patternfly/patternfly@4.159.1
├─┬ @patternfly/react-core@4.175.4
│ ├─┬ @patternfly/react-icons@4.43.15
│ │ ├── react-dom@17.0.2 deduped
│ │ └── react@17.0.2 deduped
│ ├── @patternfly/react-styles@4.42.15
│ ├── @patternfly/react-tokens@4.44.15
...
```

react-core has dependency over the rest of the react-* packages. These would be nice to be grouped in a PR. It looks like dependabot, is not grouping these (seems it just groups packages with it's peerDependencies), but that is also not very problematic, as long as the versions that exist in the package.json are compatible with each other.

So I think, I will still give it a try in the anaconda/webui repo, see how it goes and report in the future.

martinpitt:
@KKoukiou: Interesting -- we known from (painful) experience that the `@patternfly/react-*` packages often depend on each other. But I wasn't aware of `@p/patternfly` being independent. If/as we only use the react bits, can/should we perhaps drop that dependency altogether?

KKoukiou:
@martinpitt we use the plain CSS from @patternfly/patternfly in many places, and most importantly for setting the base styles, https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/patternfly/patternfly-4-cockpit.scss

That's some generic rules, that create a white-board for all the per-component CSS to step on.

martinpitt:
dependabot can finally group updates! https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/ So that's worth considering again.

jelly:
Hmm what about our custom logic for PF alpha's, can we replicate that?

```
  58   │ def prefix(name, version):
  59   │     if not version[0].isdigit():
  60   │         return version
  61   │     if name.startswith("@patternfly"):
  62   │         return "alpha"
  63   │     return "~" + version
```

jelly:
Just gave this a try in my own fork:

dependabot file https://github.com/jelly/starter-kit/blob/main/.github/dependabot.yml

PR's: https://github.com/jelly/starter-kit/pulls

It seems I basically grouped two things together and did it wrong :) But we'll it's a first trial run.


Just for myself, to force re-trigger a scan and opening PR's 
https://stackoverflow.com/questions/76782871/force-dependabot-to-scan-my-github-repo-without-open-prs
https://github.com/jelly/starter-kit/network/updates

jelly:
Let's give this a try in starter-kit, as this requires no update of `node-modules`. For cockpit-podman/cockpit-machines/cockpit we would need to figure out a solution for that

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
martinpitt, line 6: Daily feels a little aggressive, but we can give it a try..

.github/workflows/npm-update.yml
```diff
@@ -1,30 +0,0 @@
-name: npm-update
```
martinpitt, line 1: Please drop npm-update-pf.yml as well then.

martinpitt (changes requested):
Thanks! I'm all for trying this, and sorting out possible permission errors on starter-kit first -- like, is our CI properly triggered and such.

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    groups:
+      eslint:
+        patterns:
+          - "esbuild*"
```
allisonkarlitskaya, line 10: copy/paste error?

.github/workflows/npm-update.yml
jelly: Woops, sorry.

.github/dependabot.yml
jelly: Is it? The upside is that updates are more spread out, but this could also be annoying.

martinpitt (approved):
Thanks! :robot:

PR #528, opened 2022-02-11
Bump up eslint-config-standard-react and add fix for the new warnings

KKoukiou:
This started giving errors like:
  22:10  error  'Alert' is defined but never used      no-unused-vars

Use eslint-plugin-react to improve the JSX usage, not just to silence this error.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,7 @@
         "browser": true,
         "es6": true
     },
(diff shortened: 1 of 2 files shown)


KKoukiou:
I just hit this one in anaconda webui and fixed. Thought I update it here as well, so as to save these 5 mins from the person trying to update this dependency in the future.

martinpitt (approved):
Epharisto, nice!

PR #529, opened 2022-02-11
Install eslint-plugin-standard-jsx to align with what main cockpit repo does

KKoukiou:


KKoukiou:
@martinpitt sorry for unnecessary noise, this addresses the same issue with https://github.com/cockpit-project/starter-kit/pull/528 you just merged, but Cockpit already does this another way.
Not important, but also cool if we can remain consistent across repos, so sent a followp.

.eslintrc.json
```diff
@@ -4,7 +4,7 @@
         "browser": true,
         "es6": true
     },
-    "extends": ["eslint:recommended", "standard", "standard-react", "plugin:react/recommended"],
+    "extends": ["eslint:recommended", "standard", "standard-jsx", "standard-react"],
```
martinpitt, line 7: Being consistent across projects is good of course, I am just not sure which one is the "better" one. E.g.https://www.npmjs.com/package/eslint-config-standard-react says "This module is for advanced users. You probably want to use [standard](https://github.com/standard/standard) instead :)" :thinking:

KKoukiou: Hmm but the standard does not have the logic for parsing JSX code. In addition we already use standard-react which has this warning you note above. https://github.com/standard/eslint-config-standard-react

To me it makes more sense to use these eslint shareable config, than adding the plugin rules.

martinpitt (approved):
Cheers!

PR #535, opened 2022-03-08
spec: drop the `debug_package` %define

ptoscano:
The generated RPM is noarch, so by definition has no binaries to extract
debug symbols from. Hence, drop the %define that makes RPM not error out
in case there are no debug symbols around (usually in archful binaries).


martinpitt (approved):
Grazie!

PR #537, opened 2022-03-09
po: Update German translations

martinpitt:
This validates the i18n for the AppStream metadata introduced in commit
4377c4fe19.

---

Done by `make update-po` and translating the new sentences. The generated file (`make install DESTDIR=/tmp/x`) does use them, but it looks a little funny:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<component type="addon">
  <id>org.cockpit_project.starter_kit</id>
  <metadata_license>CC0-1.0</metadata_license>
  <name>Starter Kit</name>
  <name xml:lang="de">Bausatz</name>
  <summary>Scaffolding for a cockpit module</summary>
  <summary xml:lang="de">Gerüst für ein Cockpit-Modul</summary>
  <description>
    <p>
      Scaffolding for a cockpit module.
    </p>
    <p xml:lang="de">Gerüst für ein Cockpit-Modul.</p>
  </description>
  <extends>org.cockpit_project.cockpit</extends>
  <launchable type="cockpit-manifest">starter-kit</launchable>
</component>
```

I.e. `<name>` and `<summary>` look fine, but shouldn't it be `<description xml:lang="de">` with a single `<p>`? inside? This structure looks wrong.

@ptoscano FYI

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -4,7 +4,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: starter-kit 1.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-08-29 00:14+0200\n"
+"POT-Creation-Date: 2022-03-09 16:09+0100\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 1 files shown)


ptoscano:
> shouldn't it be `<description xml:lang="de">` with a single `<p>`? inside?

Translations are applied to each tag that has translatable string as text, not as "top-level" tags. Otherwise the structure inside could be different between untranslated and translated content. (At least it's my recollection from very old info on working on i18n for XML.)

The result looks OK, and in line with existing AppStream metadata files. Obviously I cannot judge the German texts, as I don't know it. :)

martinpitt:
Indeed, e.g. /usr/share/metainfo/eog.appdata.xml looks exactly the same. So, all good.

marusak (approved):
Danke!

PR #538, opened 2022-03-10
Makefile: factorize AppStream filename in own variable

ptoscano:
Makes it easier to read, and also to modify for modules outside the
cockpit-project org.


martinpitt (approved):
Grazie!

PR #541, opened 2022-03-16
Makefile: factorize cockpit.git version in own variable

ptoscano:
Use a single variable to represent the version of cockpit.git we
checkout for both the test API & the PF/React/build library; even if
they are different bits, it makes sense for projects to step both at the
same time, and avoid confusions due to different versions.


martinpitt:
This is particularly important for downstream gating tests, where we may have to bump the testlib without changing the built webpack/rpm.

ptoscano:
Oh OK. I'm happy to drop this PR, then.

Thanks for the review!

PR #542, opened 2022-03-17
Makefile: allow setup of integration tests w/o running them

ptoscano:
Add a convenience target in Makefile to perform all the setup targets
for integration tests without actually running them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ the VM, possibly with extra options for tracing and halting on test failures
 
     TEST_OS=centos-8-stream test/check-application -tvs
 
+It is possible to setup the test environment without running the tests:
+
+    TEST_OS=centos-8-stream make prepare-check
+
 You can also run the test against a different Cockpit image, for example:
 
     TEST_OS=fedora-34 make check
--- a/Makefile
+++ b/Makefile
@@ -138,9 +138,13 @@ $(VM_IMAGE): $(TARFILE) $(NODE_CACHE) bots
 vm: $(VM_IMAGE)
 	echo $(VM_IMAGE)
 
+# convenience target to setup all the bits needed for the integration tests
+# without actually running them
+prepare-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
+
 # run the browser integration tests; skip check for SELinux denials
 # this will run all tests/check-* and format them as TAP
-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
+check: prepare-check
 	TEST_AUDIT_NO_SELINUX=1 test/common/run-tests
 
 # checkout Cockpit's bots for standard test VM images and API to launch them


test/run
```diff
@@ -5,4 +5,12 @@ set -eu
 # arguments but with an appropriate $TEST_OS, and optionally $TEST_SCENARIO
 
 [ -z "${TEST_SCENARIO:-}" ] || export TEST_BROWSER="$TEST_SCENARIO"
+
+# In case this is called with "prepare" as argument, only setup the test
+# environment without running the tests
+if [ x"${1:-}" = "xprepare" ]; then
```
martinpitt, line 11: Can we please drop this? test/run is not an entry point for humans (just for CI), so it duplicates functionality which never gets exercised anywhere.

README.md
```diff
@@ -82,6 +82,13 @@ the VM, possibly with extra options for tracing and halting on test failures
 
     TEST_OS=centos-8-stream test/check-application -tvs
 
+It is possible to setup the test environment without running the tests:
+
+    export TEST_OS=centos-8-stream
+    test/run prepare
+    # as alternative to the above command:
+    make prepare-check
```
martinpitt, line 90: and drop this as well.

README.md
```diff
@@ -82,6 +82,13 @@ the VM, possibly with extra options for tracing and halting on test failures
 
     TEST_OS=centos-8-stream test/check-application -tvs
 
+It is possible to setup the test environment without running the tests:
+
+    export TEST_OS=centos-8-stream
+    test/run prepare
```
martinpitt, line 88: We usually write our docs with setting the env just for the single command. If users know what they are doing, they will know `export`, but we don't recommend that -- one tends to forget, and it's rather confusing.

martinpitt (changes requested):
Thanks! I like the idea, but let's clean it up a bit.

martinpitt:
I'm confused by the red packit/rawhide result -- the [page](https://artifacts.dev.testing-farm.io/c76133d1-6fb2-4835-8ccc-0a68afcfbf4c/) says it's still in progress?

image

image

@TomasTomecek is that a known issue?

@thrix: Can we please have "in progress" header be yellow? a green one is just too confusing at first glance, it looks like the "success" one. Yellow would match up the colors in GitHub.

martinpitt (approved):
Grazie! This does not look super-urgent, so I'd like to give packit/TF a chance to look at that odd rawhide run before I merge it. Is it ok to merge this afternoon?

ptoscano:
> Is it ok to merge this afternoon?

Ah sure. This (like the other PRs for starter-kit) are minor things I find/fix/change while trying to adapt starter-kit for subscription-manager-cockpit.

martinpitt:
Ah, the test [actually failed](https://artifacts.dev.testing-farm.io/c76133d1-6fb2-4835-8ccc-0a68afcfbf4c/work-alleA_O6p/log.txt), presumably the new cockpit version from yesterday changes the IDs around a bit. This is unrelated to this PR of course, but we'll need to adjust it. So the bug is on the TF side, showing the log as "in progress" instead of failed, and not showing the complete log.

martinpitt:
The failure with cockpit 265 is an actual regression from https://github.com/cockpit-project/cockpit/pull/16930 -- as the shell does not include all po files any more, it does not show translations of package manifest names any more. We need a better fix for that, but for now let's just revert that part? @allisonkarlitskaya  FYI

This is easy enough to reproduce on F35/F36 as well by installing the 265 update from bodhi into our test VM.

martinpitt:
I sent https://github.com/cockpit-project/cockpit/pull/17145 to fix the regression. We should do a point release after that lands.

TomasTomecek:
@martinpitt well, TF definitely told Packit that the test failed based on the result [here](https://dashboard.packit.dev/results/testing-farm/66372), so this seems like a bug in TF - maybe the test got stuck and the logs & state were not updated properly in the TF frontend?

I reported it over here: https://gitlab.com/testing-farm/general/-/issues/40

This could be a problem on our side, I didn't check logs or investigate thoroughly - just my assumption as look at the current data.

martinpitt:
@TomasTomecek : Right, as I mentioned [here](https://github.com/cockpit-project/starter-kit/pull/542#issuecomment-1070387234) this is a TF bug, and you are off the hook. At first sight it could just be another instance of https://issues.redhat.com/browse/TFT-894 but I'll let the TF people judge this. I'll point this out in your ticket.

So I think we don't need to keep this PR open any more.

PR #544, opened 2022-03-18
Makefile: depend on test/vm.install for the VM image target

ptoscano:
Make sure to recreate the VM image in case the test/vm.install script
changes.


martinpitt (approved):
Good one!

PR #545, opened 2022-03-22
Makefile: set reproducible args for tar

ptoscano:
When creating the tarballs for sources and node modules, apply some
hardening to the tar compression to remove some variances that may cause
the tarballs to not be reproducible; in particular:
- set fixed modes, and ownership
- set the mtime from the author's timestamp of the top-level commit
- sort the files by name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ NODE_MODULES_TEST=package-lock.json
 WEBPACK_TEST=dist/manifest.json
 # one example file in src/lib to check if it was already checked out
 LIB_TEST=src/lib/cockpit-po-plugin.js
+# common arguments for tar, mostly to make the generated tarballs reproducible
+TAR_ARGS = --sort=name --mtime "@$(shell git show --no-patch --format='%at')" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0
 
 all: $(WEBPACK_TEST)
 
@@ -102,12 +104,12 @@ $(TARFILE): $(WEBPACK_TEST) $(SPEC)
 	if type appstream-util >/dev/null 2>&1; then appstream-util validate-relax --nonet *.metainfo.xml; fi
 	touch -r package.json $(NODE_MODULES_TEST)
 	touch dist/*
-	tar --xz -cf $(TARFILE) --transform 's,^,$(RPM_NAME)/,' \
+	tar --xz $(TAR_ARGS) -cf $(TARFILE) --transform 's,^,$(RPM_NAME)/,' \
 		--exclude packaging/$(SPEC).in --exclude node_modules \
 		$$(git ls-files) src/lib package-lock.json $(SPEC) dist/
 
 $(NODE_CACHE): $(NODE_MODULES_TEST)
-	tar --xz -cf $@ node_modules
+	tar --xz $(TAR_ARGS) -cf $@ node_modules
 
 node-cache: $(NODE_CACHE)
 


ptoscano:
PS: I haven't tried to do a reproducible build with this; regardless, this should be a good improvement towards that goal.

Makefile
```diff
@@ -98,16 +100,16 @@ dist: $(TARFILE)
 # pre-built dist/ (so it's not necessary) and ship packge-lock.json (so that
 # node_modules/ can be reconstructed if necessary)
 $(TARFILE): export NODE_ENV=production
-$(TARFILE): $(WEBPACK_TEST) $(SPEC)
+$(TARFILE): $(WEBPACK_TEST) $(SPEC) package.json
```
martinpitt, line 103: What's that new dependency? Shouldn't that already be covered through the transitive WEBPACK_TEST → package.json dependency? If that is relevant, it should be its own separate commit.

martinpitt (dismissed):
Cheers! Asking @allisonkarlitskaya for review, as that is right up her alley.

Makefile
ptoscano: It's because of `--mtime ./package.json`, which sets the mtime for all the files. This way, if package.json changes, the tarballs are recreated (it is also mentioned in the description).

martinpitt: Ack. So this is more about being explicit, than actually changing behaviour? Because it already works that way through the WEBPACK_TEST dependency:
```
❱❱❱ make dist
cockpit-starter-kit-1.tar.xz
(Nothing gets rebuilt)

❱❱❱ touch package.json; make dist
# if it exists already, npm install won't update it; force that so that we always get up-to-date packages
rm -f package-lock.json
# unset NODE_ENV, skips devDependencies otherwise
env -u NODE_ENV npm install
[...]
NODE_ENV= node_modules/.bin/webpack
[...]
```

ptoscano: Yes, it seemed fair to do so (and I actually haven't checked whether there were other dependencies).

allisonkarlitskaya (commented):
I wrote a very long review here before, and GitHub just ate it :(

I think the tl;dr of that was this: what are you trying to accomplish here?

In particular, basing the mtime on a file that gets randomly checked out of git means that checking out starter-kit, building a tarball, erasing the checkout and doing it all over again will get two different `.tar` files, no?  That's not really my definition of reproducible.

Makefile
```diff
@@ -17,6 +17,8 @@ NODE_MODULES_TEST=package-lock.json
 WEBPACK_TEST=dist/manifest.json
 # one example file in src/lib to check if it was already checked out
 LIB_TEST=src/lib/cockpit-po-plugin.js
+# common arguments for tar, mostly to make the generated tarballs reproducible
+TAR_ARGS = --sort=name --mtime "@$(shell git show --no-patch --format='%at')" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0
```
martinpitt, line 21: I was a bit concerned about this being at the top of the file, as the Makefile is also relevant for running in release tarballs (where there is no git dir).

But it works for the basic cases:
```
❱❱❱ make dist
❱❱❱ tar -C /tmp/ -xf cockpit-starter-kit-1.tar.xz
❱❱❱ make -C /tmp/cockpit-starter-kit/ bots
make: Entering directory '/tmp/cockpit-starter-kit'
git clone --quiet --reference-if-able ${XDG_CACHE_HOME:-$HOME/.cache}/cockpit-project/bots https://github.com/cockpit-project/bots.git
info: Could not add alternate for '/var/home/martin/.cache/cockpit-project/bots': path '/var/home/martin/.cache/cockpit-project/bots' does not exist
if [ -n "$COCKPIT_BOTS_REF" ]; then git -C bots fetch --quiet --depth=1 origin "$COCKPIT_BOTS_REF"; git -C bots checkout --quiet FETCH_HEAD; fi
checked out bots/ ref 3930f8dfef4ab7454b3d45ca943087a5fe09e8c7
make: Leaving directory '/tmp/cockpit-starter-kit'
```

However, it does fail if it somehow involves rebuilding the tarball:
```
❱❱❱ make -C /tmp/cockpit-starter-kit/ rpm
make: Entering directory '/tmp/cockpit-starter-kit'
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
if type appstream-util >/dev/null 2>&1; then appstream-util validate-relax --nonet *.metainfo.xml; fi
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
tar --xz --sort=name --mtime "@" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0 -cf cockpit-starter-kit-node-1.tar.xz node_modules
tar: Substituting -9223372036854775807 for unknown date format ‘@’
tar: node_modules: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
make: *** [Makefile:112: cockpit-starter-kit-node-1.tar.xz] Error 2
make: *** Waiting for unfinished jobs....
org.cockpit-project.starter-kit.metainfo.xml: OK
touch -r package.json package-lock.json
touch dist/*
tar --xz --sort=name --mtime "@" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0 -cf cockpit-starter-kit-1.tar.xz --transform 's,^,cockpit-starter-kit/,' \
	--exclude packaging/cockpit-starter-kit.spec.in --exclude node_modules \
	$(git ls-files) src/lib package-lock.json cockpit-starter-kit.spec dist/
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
tar: Substituting -9223372036854775807 for unknown date format ‘@’
make: Leaving directory '/tmp/cockpit-starter-kit'
```

However, this also fails on main, as that already tries to prod git for information.

So this seems good enough.

martinpitt (approved):
Grazie!

PR #546, opened 2022-03-22
Makefile: add convenience 'srpm' target

ptoscano:
Add a simple 'srpm' target that builds the src.rpm file for the current
project.


ptoscano:
A possible idea I had was to change the 'rpm' target to depend on the 'srpm' target, and build the binaries from the src.rpm rather than the spec file. Since both cases are mere convenience for contributors, I think the KISS approach here should be fine.

martinpitt:
We dropped `srpm` in commit b2662eeebdb623395ca7d6ccacdd1c2746db7c42 as nobody of us considered it useful, and our CI was not using it either. If it's useful for you, I don't mind at all bringing it back. But let's please use the simpler form from the old commit?

martinpitt (approved):
Thanks! Triggering tests. I sent you an invite for the contributors team, so that you can run tests and rebuild images yourself.

PR #547, opened 2022-03-23
po: ignore js/jsx/html files under src/lib

ptoscano:
Cockpit's PF/React/build library (`pkg/lib` from cockpit.git) is checked
out as src/lib, so exclude it from the search of js/jsx, and html
sources for the extraction of i18n messages.

This avoids the cockpit messages in case the message extraction is
performed in a source tree after build.


martinpitt (approved):
Grazie!

PR #548, opened 2022-03-30
Makefile: add convenience 'print-vm' target

ptoscano:
Add a simple convenience target that prints the filename of the test
image without creating it beforehand. This can be useful to know what is
the expected filename, so it is possible to provide an own test image
instead of the generated one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,10 @@ $(VM_IMAGE): $(TARFILE) $(NODE_CACHE) bots test/vm.install
 vm: $(VM_IMAGE)
 	echo $(VM_IMAGE)
 
+# convenience target to print the filename of the test image
+print-vm:
+	echo $(VM_IMAGE)
+
 # convenience target to setup all the bits needed for the integration tests
 # without actually running them
 prepare-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -147,6 +147,10 @@ $(VM_IMAGE): $(TARFILE) $(NODE_CACHE) bots test/vm.install
 vm: $(VM_IMAGE)
 	echo $(VM_IMAGE)
 
+# convenience target to print the filename of the test image
+print-vm:
```
martinpitt, line 151: Please add this to `.PHONY`.

martinpitt (changes requested):
Hmkay -- OOI, for what do you want to use this? We sometimes run tests against running custom machines, but for that there's `--machine`/`--browser`.

ptoscano:
> Hmkay -- OOI, for what do you want to use this? We sometimes run tests against running custom machines, but for that there's `--machine`/`--browser`.

I saw that: replacing a machine is a bit more convoluted, and requires more manual setup which could be a bit too much especially in simple/personal testing.

My goal is to simply switch the base test image with an own one; something along the lines of:
```bash
export TEST_OS=...
image=$(make print-vm)
mkdir -p $(dirname $image)
cp my-own-image $image  # or qemu-img to create an overlay on the real one, etc
make check
```

martinpitt (approved):
Thanks!

PR #549, opened 2022-03-30
Makefile: add convenience 'devel-uninstall' target

ptoscano:
Add a simple convenience target to "undo" what the devel-install did
(creating a symlink to `~/.local/share/cockpit`), i.e. simply removing
that symlink.

Suggested by: Christopher Snyder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,11 @@ devel-install: $(WEBPACK_TEST)
 	mkdir -p ~/.local/share/cockpit
 	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
 
+# assumes that there was symlink set up using the above devel-install target,
+# and removes it
+devel-uninstall:
+	rm -f ~/.local/share/cockpit/$(PACKAGE_NAME)
+
 print-version:
 	@echo "$(VERSION)"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -89,6 +89,11 @@ devel-install: $(WEBPACK_TEST)
 	mkdir -p ~/.local/share/cockpit
 	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
 
+# assumes that there was symlink set up using the above devel-install target,
+# and removes it
+devel-uninstall:
```
martinpitt, line 94: Please add to `.PHONY`.

martinpitt (changes requested):
While you are at it, it would be nice to mention devel-{,un}install in README. [podman's docs](https://github.com/cockpit-project/cockpit-podman/blob/main/HACKING.md#running-out-of-git-checkout) already do this. If not, I'm happy to do that separately as a followup.

Thanks!

ptoscano:
> While you are at it, it would be nice to mention devel-{,un}install in README.

Added as extra commit - is that OK?

martinpitt (approved):
Grazie!

PR #551, opened 2022-04-04
metainfo: add few more URLs/references

ptoscano:
Even if this module is a template, add a couple more of keys to the
AppStream XML file; this way other projects can get few more hints.


martinpitt (approved):
Good idea, grazie!

PR #552, opened 2022-04-05
workflows: Drop livetest

martinpitt:
We have not used that in a long time, and it stopped working a while
ago.


.github/workflows/livetest.yml.disabled
```diff
@@ -1,118 +0,0 @@
...
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: 'Cockpit preview: https://${{ env.RELAY_HOST }}:${{ steps.remote_port.outputs.SSH_GATEWAY_PORT }}'
-            })
-
-      - name: Create an SSH private key from the secret
-        run: |
-          sudo touch /run/SINK_SSHRELAY_PRIVATE && sudo chown $(id -u):$(id -g) /run/SINK_SSHRELAY_PRIVATE
-          echo "${{ secrets.SINK_SSHRELAY_PRIVATE }}" > /run/SINK_SSHRELAY_PRIVATE
```
martinpitt, line 106: If we ever bring this back, let's use a deploy key instead of an org-wide secrets. The former are much easier to handle and refresh.

PR #554, opened 2022-04-09
package.json: bump eslint to ^8.0.0 and all eslint-x dependencies

KKoukiou:
So as to satisfy the new eslint peer dependency.

In addition drop eslint-config-standard-react as it does not release
anymore, and eslint 8.x is not compatible with the available last
release.
https://github.com/standard/eslint-config-standard-react/pull/69

Replace it with https://github.com/facebook/create-react-app#readme
which is actively maintained.

Lastly drop the deprecated
https://github.com/standard/eslint-plugin-standard


KKoukiou:
So at to keep starter-kit as template project, let's fix here possible issues and then apply to external projects as https://github.com/rhinstaller/anaconda/pull/4035#issuecomment-1093735285

martinpitt:
Fails to build in CentOS 8 stream:
```
DEBUG: NODE_ENV=production node_modules/.bin/webpack
DEBUG: ERROR in Module.createRequire is not a function
DEBUG: webpack compiled with 1 error
```
... too old node.js?

Let's fix that by not running eslint during RPM build -- it doesn't do anything good there, just waste time.

martinpitt (commented):
Many thanks! I was stumbling over this several times, in some naïve attempts to update to eslint 8. Glad to apply this to our other projects!

LGTM, but as I added a second commit this now needs to be reviewed by someone else.

marusak (approved):
Thanks!

PR #555, opened 2022-04-11
Makefile: drop `$(NODE_MODULES_TEST)` dependency from `prepare-check`

ptoscano:
The existing dependency system will already do the needed build steps
(including fetching the nodejs modules) to prepare the image for
testing, so there is no need for the tests-related target to do that.

Especially when using an own test image the assumption is that the
cockpit plugin is already installed, and thus there is no need to fetch
the nodejs modules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ print-vm:
 
 # convenience target to setup all the bits needed for the integration tests
 # without actually running them
-prepare-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
+prepare-check: $(VM_IMAGE) test/common
 
 # run the browser integration tests; skip check for SELinux denials
 # this will run all tests/check-* and format them as TAP


Makefile
```diff
@@ -158,7 +158,7 @@ print-vm:
 
 # convenience target to setup all the bits needed for the integration tests
 # without actually running them
-prepare-check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
+prepare-check: $(VM_IMAGE) test/common
```
martinpitt, line 161: But without node_modules/ you can't run tests - they need at least `chrome-remote-interface` and `sizzle`. It *happens* to work with starter-kit, as preparing the VM always builds the webpack. But this is not true for cockpit-machines, which uses a `dist/` cache and thus running webpack for rpm/VM preparation usually does *not* happen. And we want to extend the caching to other projects at some point. But aside from that happenstance, it is still conceptually true that node_modules/ must exist for running the test.

So at least you need to move the dependency to the `check:` target, but I think given what you want to do with `prepare-check` it is required there as well.

martinpitt (changes requested):

Makefile
ptoscano: Hm I see, happy to drop this then.

PR #564, opened 2022-04-21
webpack: Always regenerate and clean up dist/

### LINKED ISSUE #563: webpack does not update dist/ time stamps, causing unnecessary rebuilds
There is an issue in the dependencies of the current `Makefile`, which triggers a never-ending run of webpack.

In a clean checkout, run the following steps:
```bash
$ make prepare-check 
# if it exists already, npm install won't update it; force that so that we always get up-to-date packages
rm -f package-lock.json
[...]
From https://github.com/cockpit-project/cockpit
 * tag               267        -> FETCH_HEAD
$ make prepare-check 
make: Nothing to be done for 'prepare-check'.
```
Now let's brick it by simply simulating a change in one of the javascript sources:
```
$ touch src/app.jsx 
$ make prepare-check 
NODE_ENV=production node_modules/.bin/webpack
webpack compiled successfully
$ make prepare-check 
NODE_ENV=production node_modules/.bin/webpack
webpack compiled successfully
[... and so on ...]
```

martinpitt:
By default, webpack does not clean up dist/, so when you e.g. touch the
code and rebuild with `NODE_ENV=devel`, the old compressed assets from
a previous `production` build were left in place, and cockpit could
serve the old wrong ones. Enable the `clean` output option to fix this [1].

Also force the rewrite of all files with `compareBeforeEmit: false`, so
that the time stamps actually get updated on a webpack run. This fixes
`make` rules and unnecessary rebuilds when nothing changed.

Fixes #563

[1] https://webpack.js.org/configuration/output/#outputclean

----

See the linked issue for reproducer, but the simplest one is:
```
make
touch src/app.jsx
make  # should rebuild
make  # should do nothing
```

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -63,6 +63,8 @@ module.exports = {
     externals: { "cockpit": "cockpit" },
     devtool: "source-map",
     stats: "errors-warnings",
+    // always regenerate dist/, so make rules work
+    output: { clean: true, compareBeforeEmit: false },
 
     optimization: {
         minimize: production,


marusak (approved):
Thanks!

PR #565, opened 2022-04-21
Add basic typescript integration

isidornygren:
Adds settings for ESLint and webpack to build typescript (.ts, .tsx) files. Building the project should work just as before.

The type declarations in `src/types/cockpit.d.ts` are very basic and should preferably not be used, I just added them to at least have _some_ types for the library.

### DIFF
--- a/.babelrc.json
+++ b/.babelrc.json
@@ -9,6 +9,7 @@
                 "opera": "44"
             }
         }],
+        "@babel/preset-typescript",
         "@babel/preset-react"
     ]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,7 @@
         "browser": true,
         "es6": true
     },
-    "extends": ["eslint:recommended", "standard", "standard-jsx", "react-app"],
+    "extends": ["eslint:recommended", "standard", "standard-jsx", "react-app", "plugin:@typescript-eslint/recommended"],
     "parser": "@babel/eslint-parser",
     "parserOptions": {
         "ecmaVersion": "7",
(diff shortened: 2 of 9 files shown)


martinpitt:
Thanks At the moment this breaks tests rather thoroughly :cry: Note that the cockpit.ts annotation does not really belong here into starter-kit, it should live in cockpit itself. Someone originally attempted this in https://github.com/cockpit-project/cockpit/pull/13214 , but did not get that far. But it's already a whole lot more than the stub here :grin: 

How much does .ts make sense without a typescript annotation for cockpit.js? Does this have to be strictly hieararchical, i.e. dependencies first? Or can it just ignore all the `cockpit.*()` calls?

isidornygren:
> Thanks At the moment this breaks tests rather thoroughly 😢 Note that the cockpit.ts annotation does not really belong here into starter-kit, it should live in cockpit itself. Someone originally attempted this in [cockpit-project/cockpit#13214](https://github.com/cockpit-project/cockpit/pull/13214) , but did not get that far. But it's already a whole lot more than the stub here 😁
> 
> How much does .ts make sense without a typescript annotation for cockpit.js? Does this have to be strictly hieararchical, i.e. dependencies first? Or can it just ignore all the `cockpit.*()` calls?

I agree that the types for cockpit should live in its own project repo. I think it should be possible - although not ideal - to have an untyped library, so that "typings" file should probably be removed 😅 .

I'm a bit unsure about the tests though as I haven't been able to run them locally, and I'm not sure what's going on from reading the test details from the failed tests. If you could push me in the right direction to what the error might be / how to debug the tests I would be happy to fix whatever's happening.

martinpitt:
This is very obsolete. Cockpit has had typescript integration, several typescript libraries, and cockpit.js types for a while now. I suppose the next thing would be to convert starter-kit to full typescript.

martinpitt:
@isidornygren FRY, yesterday's PR #917 converted starter-kit to TS.

PR #568, opened 2022-04-27
`Makefile`: avoid spurious echo

ptoscano:
Disable the printing of the echo commands that write the filename of the
VM image for tests in the `vm` and `print-vm` targets.

This makes it possible to use their outputs, especially the `print-vm`
target.


martinpitt (approved):
Thanks!

PR #575, opened 2022-05-12
Fedora packaging fixes v2

ptoscano:
See https://bugzilla.redhat.com/show_bug.cgi?id=2081432#c3.

Followup of #573.


martinpitt (approved):
Very nice, thank you @Conan-Kudo and @ptoscano !

PR #582, opened 2022-06-07
Move release from cockpituous to action-release and packit

martinpitt:
This gets rid of our "release" environment with high-profile secrets,
and the cockpituous/release infra.

The upstream release uses our shared
https://github.com/cockpit-project/action-release/ action.


martinpitt:
> Do we want to release this, some day? :)

Never starter-kit, but it should provide an up to date and tested template how to do it for forked projects.

PR #595, opened 2022-07-19
Add CSS linting through stylelint

jelly:
Include CSS linting by default for our CSS/SCSS files, originally
introduced in cockpit. The stylelint configuration is copied from
760a4628282e02fbcef3.


jelly:
> I'm a bit confused -- how does this get so far without a corresponding node_modules/ commit?

This doesn't use a `node_modules` submodule git right?

PR #596, opened 2022-07-20
i18n test, maintenance

martinpitt:
This was triggered by https://bugzilla.redhat.com/show_bug.cgi?id=2077759 . It works in starter-kit, I am just admittedly not sure *why*.


marusak (approved):
Thanks!

PR #605, opened 2022-09-14
Install into /usr/local/ by default

martinpitt:
/usr is package manager territory, and not even writable on OSTree based systems such as Fedora CoreOS or RHEL Edge. This has been common and good practice in autotools for decades.

You can still override this with `make install PREFIX=...`. Use that in the RPM spec file.

---

Adapted from https://github.com/cockpit-project/cockpit-ostree/pull/281

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ export TEST_OS
 TARFILE=$(RPM_NAME)-$(VERSION).tar.xz
 NODE_CACHE=$(RPM_NAME)-node-$(VERSION).tar.xz
 SPEC=$(RPM_NAME).spec
+PREFIX ?= /usr/local
 APPSTREAMFILE=org.cockpit-project.$(PACKAGE_NAME).metainfo.xml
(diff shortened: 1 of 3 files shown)


marusak (approved):
Danke!

martinpitt:
@TomasTomecek , @lachmanfrantisek: Any known packit/TF issue right now? The test results never arrive (runs here should be fairly quick). The statuses links to e.g. https://dashboard.packit.dev/results/testing-farm/115332 don't show any Testing Farm link. On https://dashboard.packit.dev/jobs I do see these:

image

etc. -- but the "Test Results" UUIDs don't do anything (they are clickable, but no href). Trying e.g. https://artifacts.dev.testing-farm.io/53aefaa9-413c-43aa-9b7b-fa4816f26ece/ or /pipeline.log is a 404.

martinpitt:
Ah, TF tests fixed at last, thank you!

PR #607, opened 2022-10-31
cirrus: Move to Fedora 37

martinpitt:
There is no Fedora 35 image any more.

---

See [failed test](https://github.com/cockpit-project/starter-kit/pull/606/checks?check_run_id=9194202641) in #606


marusak (approved):
Danke :)

PR #621, opened 2023-02-01
build: add support for /pybridge scenario

jelly:
Build the Python bridge from the cockpit project and install it into the VM so we can run tests against it.


martinpitt (approved):
It's a bit saddening that we have to schlepp this into all of our projects.. but we can revert all of that once it actually lands. Thanks!

martinpitt (approved):
LGTM! Assuming green.

PR #622, opened 2023-02-08
.github: update release action workflow

jelly:
Our latest action-release workflow no longer uses a deprecated github-script version.


martinpitt (approved):

martinpitt:
Triggering required tests. For starter-kit the tests are so cheap, it usually doesn't make much sense to bother with no-test (the mental and procedural overhead of triggering the tests manually)

PR #624, opened 2023-02-14
Hacking rsync

jelly:
Follow up from https://github.com/cockpit-project/cockpit/pull/18325


jelly:
Once the text is approved this can go to machines, podman, etc.

README.md
```diff
@@ -51,6 +51,11 @@ the remote hostname.
 
     $ RSYNC=c make watch
 
+When developing against a remote host as a normal user `RSYNC_DEVEL` can be set
```
martinpitt, line 54:
```suggestion
When developing against a remote host as a normal user, `RSYNC_DEVEL` can be set
```

martinpitt (changes requested):
Thanks! Let's make this a bit easier to read, but LGTM!

README.md
jelly: Applied.

PR #626, opened 2023-02-23
Use whole path in patternfly imports

KKoukiou:
Otherwise we rely on treeshaking mechanisms to remove unused code, thus making our bundle size more vulnerable to bundler's ability to drop dead code.


martinpitt (approved):
Good that you had a script for this :stuck_out_tongue_winking_eye:

PR #628, opened 2023-03-08
README: document development dependencies for Fedora/Ubuntu

jelly:
Document what is required to get a test VM build done, for some new contributors the gettext dependency is not obvious to resolve.

See for example https://github.com/cockpit-project/cockpit-podman/pull/1218

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,19 @@
 
 Scaffolding for a [Cockpit](https://cockpit-project.org/) module.
 
+# Development dependencies
+
+On Debian/Ubuntu:
+
+    $ sudo apt install gettext nodejs npm make
+
+On Fedora:
+
+    $ sudo dnf install gettext nodejs npm make
+
+
 # Getting and building the source
 
-Make sure you have `npm` available (usually from your distribution package).
 These commands check out the source and build it into the `dist/` directory:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,9 +2,19 @@
...
 
+# Development dependencies
+
+On Debian/Ubuntu:
+
+    $ sudo apt install gettext nodejs npm make
+
+On Fedora:
+
+    $ sudo dnf install gettext nodejs npm make
```
martinpitt, line 13: No `npm` please -- we have tools/node-modules which runs this in a container.

martinpitt (changes requested):
Thanks!

README.md
martinpitt: Nonsense -- I was thinking `npm install`, but of course we also need it for `npm run`. Not strictly (we could call it directly as well), but our Makefile does `npm run watch`.

martinpitt (approved):
Thanks! Please trigger required tests. In general, no-test does not  really make sense for starter-kit.

jelly:
> Thanks! Please trigger required tests. In general, no-test does not really make sense for starter-kit.

I thought it would be good to limit CI usage

PR #629, opened 2023-03-14
Run `npm install` with `--ignore-scripts`

martinpitt:
https://github.com/cockpit-project/bots/blob/main/npm does that as well. We don't expect/depend on scripts, and this closes at least one major attack vector of `npm install` against developer machines.


martinpitt:
After this lands, let's add this to c-ostree and c-certificates.

jelly (approved):
Something funky going on in F37/F38

martinpitt:
I'm pretty sure that the F37/F38 packit copr builds are not my fault. Seems F37/F38 broke nodejs install?

PR #631, opened 2023-03-20
build: use `pip wheel` to get cockpit wheel

allisonkarlitskaya:
Now that we have pip in the tasks container, this is a lot easier, and it will allow us to drop the make-wheel script from the cockpit project.


martinpitt (changes requested):
Thanks! But please add pip to "Development dependencies" in README.

martinpitt (approved):
Well, actually -- this isn't really for human developers, just an intermediate hack until the py bridge becomes *the* bridge. So, fine!

PR #632, opened 2023-03-27
Bump Cockpit API to 288.1 + esbuild plugin updates, fix TF failures

martinpitt:
Preparation for moving to esbuild.


martinpitt:
I think this got stuck during the GitHub outage, force-pushing to tickle it

martinpitt:
Bumped again to pick up the new esbuild-common.js

PR #639, opened 2023-04-17
.github: add Cockpit library update action

jelly:
This action works similar to the npm-update workflow, it creates a PR every Thursday (usually after the Cockpit release) to update the COCKPIT_REPO_COMMIT to the latest version.

---

Test run:

 https://github.com/jelly/starter-kit/actions/runs/4722440632
https://github.com/jelly/starter-kit/pull/3

### DIFF
--- a/.github/workflows/cockpit-lib-update.yml
+++ b/.github/workflows/cockpit-lib-update.yml
@@ -1,0 +1,27 @@
+name: cockpit-lib-update
+on:
+  schedule:
+    - cron: '0 2 * * 4'
+  # can be run manually on https://github.com/cockpit-project/starter-kit/actions
+  workflow_dispatch:
+jobs:
+  cockpit-lib-update:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/cockpit-lib-update.yml
```diff
@@ -0,0 +1,27 @@
+name: cockpit-lib-update
+on:
+  schedule:
+    - cron: '0 2 * * 4'
+  # can be run manually on https://github.com/cockpit-project/starter-kit/actions
+  workflow_dispatch:
+jobs:
+  cockpit-lib-update:
+    runs-on: ubuntu-20.04
```
martinpitt, line 9: I'd rather this would be `ubuntu-latest`, which is 22.04 now. Let's please not introduce workflows on an already outdated Ubuntu version.

martinpitt (changes requested):
Thanks!

martinpitt (approved):

PR #641, opened 2023-04-19
Update Packit configuration

lbarcziova:
In case of any questions, please contact [Packit team](https://packit.dev/#contact).


martinpitt (approved):
Hmm, a bit sad that we have to do this now. This duplicates the targets list, and generally seems redundant. Why would this suddenly not be implied by `tests` any more?

Thanks!

PR #643, opened 2023-04-20
workflows: Update to ubuntu-latest, set explicit permissions

martinpitt:
I restricted our default workflow permissions now:

image

This has no influence on workflows with an explicit `permissions:` (and now they all have). Let's test all of them to make sure they still work.

 - [x] [npm-update](https://github.com/cockpit-project/starter-kit/actions/runs/4751164558)
 - [x] [npm-update-pf](https://github.com/cockpit-project/starter-kit/actions/runs/4751163656/jobs/8440024937)
 - [x] cockpit-lib-update (~blocked on https://github.com/cockpit-project/bots/pull/4677~) [successful workflow](https://github.com/cockpit-project/starter-kit/actions/runs/4752178616) produced #644

### DIFF
--- a/.github/workflows/cockpit-lib-update.yml
+++ b/.github/workflows/cockpit-lib-update.yml
@@ -7,6 +7,9 @@ on:
 jobs:
   cockpit-lib-update:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: write
     steps:
       - name: Set up dependencies
         run: |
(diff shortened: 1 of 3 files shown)


martinpitt:
Re-trying without "contents:" permissions for cockpit-lib-update. Maybe we'll need "read", but as this is a public project, let's try without first.

jelly (approved):
Danke!

PR #648, opened 2023-05-10
package.json: Pin down versions of @patternfly/react-{styles,icons}

martinpitt:
These are already installed as dependencies, and we do the same in other Cockpit projects. The latest react-styles version became incompatible with the react-core version, causing a build failure.

----

See https://github.com/cockpit-project/bots/pull/4746 and other failed image refreshes.


mvollmer (approved):
Thänks!

PR #657, opened 2023-05-25
patternfly-5-overrides should be auto-imported by the page.scss file

KKoukiou:
page.scss sould be imported by all pages.


martinpitt (approved):
Epharisto!

PR #658, opened 2023-05-26
Makefile: bump our test/common dependency

allisonkarlitskaya:
... and make use of the new pywrap feature from our test.


martinpitt (dismissed):

martinpitt (approved):
Thanks to both of you!

PR #683, opened 2023-08-01
Revert "build: add support for /pybridge scenario"

martinpitt:
We don't need this any more, the pybridge landed in all planned distributions.

Do keep the more explicit and correct handling of `$TEST_BROWSER` in test/run, though.

This reverts commit 03d02f398a4d92926e81ae92c04fa25411243004.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,22 +153,10 @@ rpm: $(TARFILE) $(NODE_CACHE) $(SPEC)
 	rm -r "`pwd`/rpmbuild"
 	rm -r "`pwd`/output" "`pwd`/build"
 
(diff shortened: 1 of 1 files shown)


jelly (approved):
This still tests centos-8-stream, which still has a [scenario here](https://github.com/cockpit-project/bots/pull/5072#issue-1831793006). But I suppose that is only for Cockpit?

martinpitt:
Yeah, we never tested starter-kit c8s/pybridge. Perhaps we want to start doing that for podman/machines, but it feels relatively uninteresting for starter-kit.

PR #684, opened 2023-08-01
package.json: drop deprecated eslint-plugin-standard

jelly:
eslint-config-standard no longer requires it since 16.0.0.


martinpitt (approved):
:broom:

PR #704, opened 2023-09-18
package.json: drop po2json

jelly:
As we did in 1e92eb6815eb48aab590790ecd677aebdaca3f7c


martinpitt (approved):
Ah nice! Presumably gettext-parser previously was a transitive dependency of po2json?

jelly:
> Ah nice! Presumably gettext-parser previously was a transitive dependency of po2json?

It was!

PR #715, opened 2023-10-02
package.json: drop flowtype plugin

jelly:
None of the Cockpit projects use flowtype's type annotation.

Follow up PR's:

* [ ] machines
* [ ] podman
* [ ] ostree
* [ ] navigator
* [ ] certificates

In Cockpit we already dropped flowtype.


martinpitt (approved):
Thanks!

PR #783, opened 2024-02-07
github: allow dependeabot to update github workflow actions

jelly:
Github usually updates their actions once a while and then warns about a node env getting deprecated in runs. Which is not super easily spotted by a developer until it's too late, so let's like npm let dependabot handle updating.

---

This is a draft for now, let's discuss this in the team meeting today.


martinpitt (approved):
We discussed that and there was consensus. Thanks!

PR #791, opened 2024-02-09
.github: don't run integration tests on actions changes

jelly:


martinpitt (approved):
Thanks! We'll have to manually trigger the required tests, but that's better than the avalanche of needless bot runs this morning.

martinpitt:
I.e. for starter-kit that isn't actually a problem, but I suppose you'll apply it to all our other projects :grin: 

Also, ironic side node: This could have been no-test.

jelly:
> I.e. for starter-kit that isn't actually a problem, but I suppose you'll apply it to all our other projects 😁

Yes, in my head it goes start in starter-kit and let it spread :)

> Also, ironic side node: This could have been no-test.

:100:

PR #797, opened 2024-02-20
workflows: Switch release to cockpit/tasks container

martinpitt:
The unit-tests container was dropped in
https://github.com/cockpit-project/cockpit/commit/f16f1fc14b88c

Also switch the template to current actions/checkout version.

---

After this lands, I'll apply it to all our other projects. Then we can drop https://ghcr.io/cockpit-project/unit-tests (which is deadware by now).

### DIFF
--- a/.github/workflows/release.yml.disabled
+++ b/.github/workflows/release.yml.disabled
@@ -11,14 +11,14 @@ jobs:
   source:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/cockpit-project/unit-tests
+      image: quay.io/cockpit/tasks:latest
       options: --user root
     permissions:
       # create GitHub release
       contents: write
     steps:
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
We already run `make distcheck` as part of our unit tests, so we know that this works, right?

This really feels "right", thanks.  It was odd that we were building our release in our "unit-tests" container...

Something to consider for the future: we have an idea for how we'll make CI run the container that we read from `.cockpit-ci/container`, but do we have a plan for github workflows?

martinpitt:
> do we have a plan for github workflows?

We could do something similar to the cockpit lib update. But honestly, I wouldn't bother for the release workflow. We will regularly update the tasks container for tests, which includes `make dist`, so building a release tarball really should work on `:latest`.

PR #812, opened 2024-03-14
test/browser: run tests inside the tasks container

allisonkarlitskaya:
This simplifies the "outside" setup quite a bit and gives us the same
tasks container that test runs on the Cockpit CI run under.
    
This change also implicitly changes the test browser to chromium (which
is no longer an issue — we just use it from the tasks container).


test/browser/run-test.sh
```diff
@@ -34,9 +36,13 @@ echo "TEST_ALLOW_JOURNAL_MESSAGES: ${TEST_ALLOW_JOURNAL_MESSAGES:-}"
 echo "TEST_AUDIT_NO_SELINUX: ${TEST_AUDIT_NO_SELINUX:-}"
 
 RC=0
-test/common/run-tests --nondestructive --machine 127.0.0.1:22 --browser 127.0.0.1:9090 $EXCLUDES || RC=$?
+./test/common/run-tests \
+    --nondestructive \
+    --machine localhost:22 \
+    --browser localhost:9090 \
```
martinpitt, line 42: Was the `--network=host` deliberate, i.e. does it simplify things significantly? Shouldn't this work equally well with `--machine _gateway:22`? Then it would isolate the test better.

martinpitt (commented):
This looks really nice, thanks for working on this! The TF seems to be slow today, let's wait for the results.

martinpitt:
> The TF seems to be slow today

ah no, you just pushed 10 mins ago, so that seems to be fine.

test/browser/run-test.sh
allisonkarlitskaya: Great idea.  This is the kind of thing that obviously exists, if you think about it, but I had never thought about it.  I'll look into it.

allisonkarlitskaya:
The aarch64 thing is kinda obvious — I've been kinda curious about building a tasks container there for a little while anyway (since we've been talking about our rasp-pi performance profiling ideas).

The fedora-39-x86_64 thing is a bit weirder.  Do we assume that /etc/cockpit exists?  And if so, why doesn't it exist in this case?

martinpitt:
@allisonkarlitskaya This is just a follow-up error from the browser crash / timeout / hard test kill before. This is Chromium using oodles of memory again, and we can neither influence the total RAM of the VM nor have a chance to spawn a new VM after a hard kill. I'd say, just use Firefox -- it has proven to work in TF and is the official RHEL browser anyway.

jelly:
> The aarch64 thing is kinda obvious — I've been kinda curious about building a tasks container there for a little while anyway (since we've been talking about our rasp-pi performance profiling ideas).
> 
> The fedora-39-x86_64 thing is a bit weirder. Do we assume that /etc/cockpit exists? And if so, why doesn't it exist in this case?

And don't forget Cockpit also build tests on powerpc/s390x (but I that is only a COPR build I think)

allisonkarlitskaya:
> And don't forget Cockpit also build tests on powerpc/s390x (but I that is only a COPR build I think)

ugh.  I did forget about that.

jelly:
> > And don't forget Cockpit also build tests on powerpc/s390x (but I that is only a COPR build I think)
> 
> ugh. I did forget about that.

Update, it's only building rpm's on s390x we don't run tests on it. But it could be that in the future TF supports a new architecture like RISC-V (But that depends on what AWS offers).

test/vm.install
```diff
@@ -1,9 +1,10 @@
 #!/bin/sh
 # image-customize script to prepare a bots VM for testing this application
 # The application package will be installed separately
-set -eu
+set -eux
```
martinpitt, line 4: Does that make the standard `make vm` more noisy?

martinpitt (approved):
Nice! I'm suspicious about using chromium here, given that we've seen that dreaded V8 OOM before. But that'll probably be more prominent when we start to test bigger projects.

jelly:
I'd prefer if we keep running a runner on Firefox so either making the TF tests run on Firefox again or adding `{DEFAULT_OS}/firefox` in bots to the testmap.

martinpitt:
@jelly Right, I sent #815.

PR #815, opened 2024-03-16
FMF test tweaks 

martinpitt:
See https://github.com/cockpit-project/starter-kit/pull/812#issuecomment-1999783563 and the changes which @allisonkarlitskaya applied to our other projects.

This also fixes TEST_OS, see https://github.com/cockpit-project/cockpit-machines/pull/1502#issuecomment-2001922204

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -36,5 +36,6 @@ exec podman \
         --security-opt=label=disable \
         --volume="${TMT_TEST_DATA}":/logs:rw,U --env=LOGS=/logs \
         --volume="$(pwd)":/source:rw,U --env=SOURCE=/source \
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya (commented):
Let's wait until after we land cockpit-machines and -podman.

For the record, I wouldn't be against bind-mounting `--volume:/:/run/host:ro` to be a bit like how toolbox does it.  I even notice that someone added the `/run/host` symlink on my *host* system, so then `. /run/host/usr/lib/os-release` would always get the info from the host machine.

martinpitt:
Hmm, I find bind-mounting the entire host to /run/host a bit gross, but I don't mind bind-mounting just os-release.

allisonkarlitskaya (approved):

PR #821, opened 2024-03-22
spec: Support SUSE specic BuildRequires

Nykseli:
In order to [openSUSE tumbleweed testing image](https://github.com/cockpit-project/bots/pull/5954) pass tests in the starter kit, the spec file needs to be updated to support it.


martinpitt (approved):
Cheers! As you tested this locally, let's just land it, then we can test the image more easily.

PR #825, opened 2024-03-26
workflows: Move tasks-container-update to Sunday night

martinpitt:
We previously ran tasks-container-update and cockpit-lib-update at the same time, which sometimes collides. Move it to the night between Sunday and Monday instead, so that we have the updates ready for us to inspect when we start our week.

### DIFF
--- a/.github/workflows/tasks-container-update.yml
+++ b/.github/workflows/tasks-container-update.yml
@@ -1,7 +1,7 @@
 name: tasks-container-update
 on:
   schedule:
-    - cron: '0 2 * * 4'
+    - cron: '0 2 * * 1'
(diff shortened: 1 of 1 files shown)


martinpitt:
@mvollmer @jelly We talked about this recently. Feel free to bikeshed about the particular time, of course :grin: After that lands, I'll send the same PR to our other projects.

jelly (approved):

jelly:
Can't think of something else that runs on Sunday so sounds good to me!

martinpitt:
@jelly Note that technically this already is Monday (Sunday is day 0), it's running on Monday 02:00. But many developers would say it's Sunday night :grin:

martinpitt:
That's why the commit message says "the night between Sunday and Monday"

PR #836, opened 2024-04-11
test: More precise dropdown selector

martinpitt:
Cockpit 315 changed the session menu from the deprecated PF Dropdown to
the new one, via cockpit-components-dropdown. `DropdownItem` now renders
the class name into *both* the `<li>` and the `<button>` inside of it,
so select the button to avoid an ambiguous selector.

----

https://artifacts.dev.testing-farm.io/b991462d-eb0f-49f6-a7b6-c950757e4199/

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -33,7 +33,7 @@ class TestApplication(testlib.MachineCase):
         b.switch_to_top()
         # the menu and dialog changed several times
         b.click("#toggle-menu")
-        b.click(".display-language-menu")
+        b.click("button.display-language-menu")
         b.wait_popup('display-language-modal')
         b.click("#display-language-modal [data-value='de-de'] button")
         b.click("#display-language-modal button.pf-m-primary")


martinpitt:
As I suspected, [this failure](https://artifacts.dev.testing-farm.io/b991462d-eb0f-49f6-a7b6-c950757e4199/) is independent of #835.

martinpitt:
The html dump from the failed test indeed shows two class matches, button in li:
```html
<li class="pf-v5-c-menu__list-item display-language-menu" role="none" data-ouia-component-type="PF5/DropdownItem" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-DropdownItem-2">
    <button tabindex="-1" class="pf-v5-c-menu__item display-language-menu" role="menuitem" type="button">
        <span class="pf-v5-c-menu__item-main"><span class="pf-v5-c-menu__item-text">Display language</span></span>
    </button>
</li>
```

I can reproduce this with updating to cockpit 315 in the VM. With 314 it was still `<li role="none">` without any further properties (in particular not a duplicate).

With [this trick](https://discourse.mozilla.org/t/add-freeze-page-in-5-seconds-button-to-inspector/86723) I can inspect the menu in an interactive Firefox.

There is really no good reason for this -- the code in https://github.com/cockpit-project/cockpit/blob/f50edd913c707a9e584cec2b74b2c073ca135064/pkg/shell/topnav.jsx#L214 adds it to the `DropdownItem`, but that should certainly not apply to both the li *and* the button!? I.e. this is a bug inside of the PF Dropdown.

jelly (approved):

PR #920, opened 2024-07-12
Read host name just once

martinpitt:
Add an empty dependency array to useEffect(), to go back to the behaviour since before commit 4bbb291281556.

Thanks to @subhoghoshX for spotting!


martinpitt:
Ah, let's use @subhoghoshX 's PR #921

PR #921, opened 2024-07-12
Add empty dependency array to useEffect hook

subhoghoshX:
This limits the file watch listener to only attach once and not on every render which is unnecessary.


martinpitt (approved):
Lol, we collided, I closed mine (#920). Thanks!

subhoghoshX:
> Lol, we collided, I closed mine (#920). Thanks!

Thanks. I should've been more clear in the comment, that I'm fixing it.

PR #926, opened 2024-07-16
CDP replacement candidate: webdriver npm module bidi demo

martinpitt:
I must say I feel right at home here. It's a bit more abstract than CDP, and the nice thing is that it does all the browser setup/process management etc. for you. No faffing around any more with frame tracking, execution context IDs, etc.

No pythonization at all yet, I just threw together a really cheap startup of the test VM -- these parts are not dependent on the particular choice of webdriver/puppeteer/etc., and are straightforward.

https://issues.redhat.com/browse/COCKPIT-1141

### DIFF
--- a/bidi.js
+++ b/bidi.js
@@ -109,8 +109,10 @@ try {
     // base64 encoded PNG, works fine
     // const screenshot = await b.takeScreenshot();
 
+    /*
     console.log("\n\nFinished. Press Control-C to exit");
     await new Promise(resolve => setTimeout(resolve, 100000));
+    */
 } finally {
     await b.deleteSession();
 }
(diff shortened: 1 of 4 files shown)


martinpitt:
Cool, green on first try!

martinpitt:
Oh, some missing things:

- [ ] This needs to go:
```
2024-07-16T16:18:27.806Z INFO webdriver: Downloading Chromedriver v126.0.6478.126
2024-07-16T16:18:28.286Z INFO webdriver: Download of Chromedriver v126.0.6478.126 was successful
```

- [ ] Run with firefox (just triggered)
- [ ] Make a screenshot

martinpitt:
I see no advantage of doing this over #934 -- plus, most of the bits are missing, and it's really hard to wean off this npm module from downloading the drivers.

PR #979, opened 2024-09-23
build.js: prefer startsWith over regex with a caret

jelly:
Found by eslint-plugin-unicorn.

---

I personally prefer this simplification, it is also present in machines (not podman/files/cockpit)


martinpitt (approved):
Oh yes, nice cleanup. Thanks to the unicorn! :unicorn:

PR #1028, opened 2024-11-08
Move to PatternFly Elements and lit

martinpitt:
PFE uses lit, so it makes sense to just use that for the main
application as well. Web components are pretty much React built into the
web platform, and lit adds some convenience around that.

This is mostly a demo -- for real Cockpit pages, PF Elements is still
missing too many components.

----

This is part of my Day of Learning  -- I don't think we should just land that. Perhaps we want to keep it in a branch, or a separate starter-kit-pfe project, I don't know yet. For now this is just a demo.

The size difference is quite impressive:
 - index.js.gz: 80 kB → 29 kB (this includes all PF CSS!)
 - index.css.gz: 22930 Bytes → disappears completely
 - node_modules: 381MB → 202 MB

The UI isn't surprising:

image

But this changed the `<Alert>` to a normal `<p>`, as PF Elements does not have an alert yet: https://github.com/patternfly/patternfly-elements/issues/2539

 - [x] https://github.com/cockpit-project/cockpit/pull/21233
 - [x] pull that in via #1035

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -29,6 +29,6 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
 </head>
 
 <body>
-    <div id="app"></div>
+    <ct-application></ct-application>
 </body>
 </html>
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,12 +4,11 @@
         "browser": true,
         "es6": true
     },
-    "extends": ["eslint:recommended", "standard", "standard-jsx", "standard-react"],
+    "extends": ["eslint:recommended", "standard"],
     "parserOptions": {
         "ecmaVersion": "2022",
         "sourceType": "module"
     },
-    "plugins": ["react", "react-hooks"],
     "rules": {
         "indent": ["error", 4,
             {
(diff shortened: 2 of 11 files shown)


martinpitt:
Ah, firefox!

bennypowers:
curious how this might affect final bundle sizes, page loading speed, and runtime performance

martinpitt:
@bennypowers bundle sizes are in the description. I didn't test performance, for starter-kit everything is so fast that it makes no practical difference.

martinpitt:
That Firefox failure on Fedora 41 is real, I can reproduce it locally. For some reason querySelectorAllDeep() "sees" old elements after a page load. At first sight there's no caching involved, but this requires some deeper debugging.

src/index.html
```diff
@@ -29,6 +29,6 @@
 </head>
 
 <body>
-    <div id="app"></div>
+    <ct-application />
```
bennypowers, line 32:
```suggestion
    <ct-application></ct-application>
```

custom elements can't be self-closing https://stackoverflow.com/questions/23961178/do-custom-elements-require-a-close-tag

src/index.tsx
```diff
@@ -17,16 +17,6 @@
...
-
-import { Application } from './app.jsx';
-
-import "patternfly/patternfly-5-cockpit.scss";
-import './app.scss';
-
-document.addEventListener("DOMContentLoaded", () => {
-    createRoot(document.getElementById("app")!).render(<Application />);
-});
+import './app.jsx';
```
bennypowers, line 22:
```suggestion
import './ct-application.js';
```

src/index.tsx
bennypowers, line 1: suggest renaming to index.ts

src/app.tsx
bennypowers, line 1: suggest rename to ct-application.ts

martinpitt:
תודה לך!

I applied your suggestions.

martinpitt:
Got rid of `sass` as well.

martinpitt:
This was a research project, no need to keep the PR open.

PR #1226, opened 2025-08-19
revive cirrus ci?

jelly:


jelly:
We'll need to create an issue [here](https://github.com/cirruslabs/cirrus-ci-docs/issues).

.cirrus.yml
```diff
@@ -13,7 +13,7 @@ test_task:
       - TEST_OS: fedora-42
       - TEST_OS: centos-9-stream
 
-  fix_kvm_script: sudo chmod 666 /dev/kvm
+  accel_check_script: sudo chmod 666 /dev/kvm
```
martinpitt, line 16: Note the "fix_kvm" or "accel_check" isn't magic at all. That's just to label the scripts. Only `_script` is magic. so this is a no-op. The thing is that sudo doesn't work any more, or /dev/kvm has wrong permissions (root:pcp 0600)

jelly:
Cirrus lab is no more https://cirruslabs.org/

PR #1230, opened 2025-08-26
cirrus: Re-enable and drop chmod

### LINKED ISSUE #1225: re-enable cirrus ci
The job failed with:

```
sudo chmod 666 /dev/kvm
sudo: PAM account management error: Authentication service cannot retrieve authentication info
sudo: a password is required
```

According to the [docs](https://cirrus-ci.org/guide/linux/#kvm-enabled-privileged-containers) this is no longer needed but we don't have `accel_check_script: emulator -accel-check`.

martinpitt:
Changing the /dev/kvm permissions is no longer necessary [1], and running `sudo` isn't allowed any more.

This reverts commit 948ce2f94e.

Fixes #1225

[1] https://cirrus-ci.org/guide/linux/#kvm-enabled-privileged-containers

----

Draft for now, most likely I'll need to iterate/debug a bit.


martinpitt:
Hmm, that's no good:
```
ls -l /dev/kvm
crw-rw---- 1 root pcp 10, 232 Aug 26 08:04 /dev/kvm
```

and without sudo we can't fix that.

PR #1248, opened 2025-09-19
git: Add generated specfiles to `.gitignore`

Venefilyn:
Otherwise they'll be added to the commit for new projects using the
starter kit

Signed-off-by: Freya Gustavsson <freya@venefilyn.se>


martinpitt (approved):
Cheers!

PR #1467, opened 2026-09-09
.github/workflows: Add issues to GH Project

### LINKED ISSUE #23714: GitHub Project for tracking issues
We've been using a Cockpit Pilot board where we add issues that we want there, it is a manual process and not something we are fully relying on and instead have as a sort of "take a look later" kinda queue.

Instead of adding things manually, if we were to add issues automatically from all our repositories and track them we can faster triage, avoid duplicate work, and better track things between ourselves and each other.

I've created an Ace Pilot Board (Name TBD) that already accepts [issues here in cockpit as well as bot-specific PRs](https://github.com/cockpit-project/cockpit/pull/23703). I've also created PRs for other repos that need to be merged for this to be closed:

- [x] https://github.com/cockpit-project/cockpit-files/pull/1487
- [x] https://github.com/cockpit-project/cockpit-machines/pull/2789
- [x] https://github.com/cockpit-project/starter-kit/pull/1467
- [x] https://github.com/cockpit-project/subscription-manager-cockpit/pull/355
- [x] https://github.com/cockpit-project/cockpit-ostree/pull/1144
- [x] https://github.com/cockpit-project/cockpit-podman/pull/2673
- [x] https://github.com/cockpit-project/cockpit-project.github.io/pull/919
- [x] https://github.com/cockpit-project/cockpit-docs/pull/14

Venefilyn:
Adds a GitHub Project workflow which auto-adds issues and adds bot specific PRs.

- One workflow as we're gonna be copying this around a lot, so we may as well make it a bit easier.  Keeps things as DRY as they can be, given the circumstances.
- Use the `gh-project` environment (there has never been a global `GH_PROJECTS` secret and we don't want to add one)
- Use `pull_request_target`: we don't get access to the environment (and secret) otherwise.  This is generally considered bad practice but it's pretty safe here since we never check out the PR's code.
- Add a comment about the 3 other approaches we tried/considered

### DIFF
--- a/.github/workflows/gh-project.yml
+++ b/.github/workflows/gh-project.yml
@@ -1,0 +1,39 @@
+# GitHub Projects lets you import from multiple repositories on creation, but the
+# built-in "Auto-add to project" workflow only supports a single repository (without
+# paying for more). This workflow replicates that functionality across all of our
+# repositories.
+#
+# We also considered:
+#   - a single workflow in a single repository (eg: bots) that checks all of
+#     the repositories periodically: this wouldn't be realtime, and it would
+#     be a lot of wasted action runs, 99% of which would do nothing at all
+#   - driving a small Python helper from our webhook: this would be a nice
+#     option, but most of our "GitHub administrativia" tasks are done via GitHub
+#     workflows, so let's keep with the trend
+
+name: Add issues and bot PRs to Cockpit GitHub Project
+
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Note to self: add starter-kit to the list in the ci-secrets repo.

